PR #61, opened 2020-09-13
Skipmap

LuisMartinSchick:
Finished development on the skipmap plugin.
Might need a revision.

### DIFF
--- a/plugins/skipmap/index.js
+++ b/plugins/skipmap/index.js
@@ -1,0 +1,74 @@
+import { CHAT_MESSAGE } from 'squad-server/events';
+
+export default {
+  name: 'skipmap',
+  description:
+    'The `skipmap` plugin will allow players to vote via `+/-` if they wish to skip the map',
+
+  defaultEnabled: true,
+  optionsSpec: {
+    command: {
+      required: true,
+      description: 'The name of the command to be used in chat.',
+      default: '!skipmap'
+    },
+
+    startTimer: {
+      required: false,
+      description: 'Time to wait before allowing the vote to go through.',
+      default: 15 * 60 * 1000
+    },
+
+    pastVoteTimer: {
+      required: false,
+      description: 'Time that needs to have passed since the last vote.',
+      default: 10 * 60 * 1000
+    },
+
+    endTimer: {
+      required: false,
+      description: 'Time before the end timer of the round where the votes are no longer valid.',
+      default: 15 * 60 * 1000
+    },
+
+    minimumVotes: {
+      required: false,
+      description: 'The minimum required amount of votes for the vote to go through',
+      default: 20
+    },
+
+    reminderAmount: {
+      required: false,
+      description: 'The amount of reminders broadcasted.',
+      default: 5
+    },
+
+    reminderInterval: {
+      required: false,
+      description: 'The time between individual reminders.',
+      default: 2 * 60 * 1000
+    }
+  },
+
+  /*
+  TODO:
+  - Check Start Conditions
+    - Start Timer
+    - End Timer
+    - No previous vote failed within past X minutes, as set by options
+  - Check Vote End Conditions
+    - Enough Votes
+  - Voting
+    - Chat listener, add/subtract from counter
+      -> If counter positive in the end, skip map
+    - Broadcast reminders, as set by options
+  -
+   */
+
+  init: async (server, options) => {
+    server.on(CHAT_MESSAGE, (info) => {
+      if (info.message.chat.startsWith(options.command)) {
+      }
+    });
+  }
+};
--- a/plugins/skipmap/index.js
+++ b/plugins/skipmap/index.js
@@ -1,4 +1,4 @@
-import { CHAT_MESSAGE } from 'squad-server/events';
+import { CHAT_MESSAGE, NEW_GAME } from 'squad-server/events';
 
 export default {
   name: 'skipmap',
@@ -11,6 +11,12 @@ export default {
       required: true,
       description: 'The name of the command to be used in chat.',
       default: '!skipmap'
+    },
+
+    voteDuration: {
+      required: false,
+      description: 'How long the vote should go.',
+      default: 5 * 60 * 1000
     },
 
     startTimer: {
@@ -37,12 +43,6 @@ export default {
       default: 20
     },
 
-    reminderAmount: {
-      required: false,
-      description: 'The amount of reminders broadcasted.',
-      default: 5
-    },
-
     reminderInterval: {
       required: false,
       description: 'The time between individual reminders.',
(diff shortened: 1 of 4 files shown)


plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
+import { CHAT_MESSAGE, NEW_GAME } from 'squad-server/events';
+
+export default {
+  name: 'skipmap',
+  description:
+    'The `skipmap` plugin will allow players to vote via `+/-` if they wish to skip the map',
+
+  defaultEnabled: true,
```
Thomas-Smyth, line 8: I'd default off... I doubt this will be a commonly used feature.

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
+import { CHAT_MESSAGE, NEW_GAME } from 'squad-server/events';
+
+export default {
+  name: 'skipmap',
+  description:
+    'The `skipmap` plugin will allow players to vote via `+/-` if they wish to skip the map',
```
Thomas-Smyth, line 6: Replace \` with `<code>...</code>`. The recent documentation changes stopped \` from working.

I'd state the "current" map. Skipping the next map may be something we want to add in the future.

LuisMartinSchick:
Forgot to clear the votes.

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+    let timeLastVote;
+    let voteActive;
+    const playerVotes = {};
+
+    server.on(CHAT_MESSAGE, (info) => {
+      // Run through conditions
+      // check if message is command
+      if (!info.message.chat.startsWith(options.command)) return;
+      // check if enough time has passed since start of round and if not, inform the player
+      if (server.layerHistory[0].time < options.startTimer) {
```
Thomas-Smyth, line 67: Not sure this is right? Usually for this we would do `server.layerHistory[0].time < Date.now() - options.startTimer`. Now it looks like you're comparing a date with a number.

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+        const min = Math.floor(
+          ((options.startTimer - server.layerHistory[0].time) % (1000 * 60 * 60)) / (1000 * 60)
+        );
+        const sec = Math.floor(
+          (options.startTimer - server.layerHistory[0].time) % ((1000 * 60) / 1000)
+        );
+        server.rcon.warn(
+          info.steamID,
+          'Not enough time has passed since the start of the match. Please try again in ' +
```
Thomas-Smyth, line 76: Might be tidier to use JavaScript's template literals to insert the variables in the string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+            min +
+            'm ' +
+            sec +
+            's'
+        );
+        return;
+      }
+      // check if enough time remains in the round, if not, inform player
+      if (server.layerHistory[0].time < 2 * 60 * 60 * 1000 - options.endTimer) {
```
Thomas-Smyth, line 85: I'd probably rework this condition a little as matches rarely last 2 hours. Instead, I'd maybe change this condition to "is it within x amount of the start time?", which would allow servers to go you can skip map providing it's between 5 and 10 minutes of the match starting using a combination of conditions. This might be more appealing as some servers may wish to block maps being changed after teams have already sunken 30 minutes of effort into the round.

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+        server.rcon.warn(info.steamID, 'Too close to expected end of match.');
+        return;
+      }
+
+      // check if enough time has passed since the last vote
+      if (timeLastVote - new Date() < options.pastVoteTimer) {
```
Thomas-Smyth, line 91: Not sure this is right again. Probably should be `timeLastVote < new Date() - options.pastVoteTimer`, i.e. is the last vote before (`<`) 5 minutes ago.

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+              votePos +
+              ' voted in favour, ' +
+              voteNeg +
+              'against.'
+          );
+        }
+      }, options.voteDuration);
+
+      // As a vote happened, stop any further votes from happening until enough time has passed
+      timeLastVote = new Date();
```
Thomas-Smyth, line 133: I'd maybe move this up into the timeout. I think it would be better to have no vote within x minutes of the last one ending as opposed to x minutes of the last on starting.

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+      if (!voteActive) return;
+
+      // Check if player has voted previously, if yes, remove their vote
+      if (playerVotes[info.steamID]) {
+        if (playerVotes[info.steamID] === '+') votePos--;
+        else voteNeg--;
+      }
+
+      // Record player vote
+      if (info.message.startsWith('+')) votePos++;
```
Thomas-Smyth, line 153: I'd probably change this to just `=== '+'` after stripping any whitespace. Otherwise the following chat could lead to an accidental upvote:
```
Tommy: Why do you want to skip a map just because you're loosing.
Morph: +1
```

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+      else if (info.message.startsWith('-')) voteNeg--;
+      playerVotes[info.steamID] = info.message;
+    });
```
Thomas-Smyth, line 156: Maybe give some feedback similar to the mapvote:
 * `You voted to (not) skip the map.`
 * `Powered by SquadJS`

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+      // check if enough time has passed since the last vote
+      if (timeLastVote - new Date() < options.pastVoteTimer) {
+        server.rcon.warn(info.steamID, 'Not enough time has passed since the last vote.');
+        return;
+      }
+
+      // Actual vote
+      voteActive = true;
+      votePos = 1;
+      playerVotes[info.steamID] = info.message;
```
Thomas-Smyth, line 99: Replace `info.message` with `+` otherwise, it will store `!skipmap` as opposed to `+`/`-` that you use later.

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+      voteNeg = 0;
+      // Set reminders
+      intervalReminderBroadcasts = setInterval(() => {
+        server.broadcast(
+          'A vote to skip the current map has started. Please vote in favour with + or against with -.'
+        );
+        server.broadcast('Current counter is: ' + votePos + ' in favour, ' + voteNeg + ' against.');
```
Thomas-Smyth, line 106:
```
      intervalReminderBroadcasts = setInterval(async () => {
        await server.broadcast(
          'A vote to skip the current map has started. Please vote in favour with + or against with -.'
        );
        await server.broadcast('Current counter is: ' + votePos + ' in favour, ' + voteNeg + ' against.');
```

We want to ensure they're done in the correct order so we should `await` the server to confirm that the broadcast has been received/made first before making the second.

plugins/skipmap/index.js
```diff
@@ -0,0 +1,158 @@
...
+
+      // As a vote happened, stop any further votes from happening until enough time has passed
+      timeLastVote = new Date();
+    });
+
+    // Clear timeouts and intervals when new game starts
+    server.on(NEW_GAME, (info) => {
+      clearInterval(intervalReminderBroadcasts);
+      clearTimeout(timeoutVote);
+    });
```
Thomas-Smyth, line 140: Don't forget to set the vote as inactive here too.

Thomas-Smyth (changes requested):

PR #78, opened 2020-09-24
commands made case-insensitive

LeventHAN:
This will fix the issue with commands being case sensitive.

### DIFF
--- a/plugins/chat-commands/index.js
+++ b/plugins/chat-commands/index.js
@@ -34,7 +34,7 @@ export default {
       // loop through all possibilities
       for (const command of options.commands) {
         // check if message is a command
-        if (!info.message.startsWith(command.command)) continue;
+        if (!info.message.toLowerCase().startsWith(command.command.toLowerCase())) continue;
         // check if ignored channel
         if (command.ignoreChats.includes(info.chat)) continue;
         // React to command with either a broadcast or a warning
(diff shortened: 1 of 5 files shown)


Thomas-Smyth (changes requested):
Missing some changes in `mapvote-123` and `mapvote-did-you-mean` for the other options, e.g. `restart`.

I think that perhaps the approach taken in team-randomizer, i.e. using regexes matching the start of the string with the `i` flag, might be more elegant: https://github.com/Thomas-Smyth/SquadJS/blob/master/plugins/team-randomizer/index.js

LeventHAN:
> Missing some changes in `mapvote-123` and `mapvote-did-you-mean` for the other options, e.g. `restart`.
> 
> I think that perhaps the approach taken in team-randomizer, i.e. using regexes matching the start of the string with the `i` flag, might be more elegant: https://github.com/Thomas-Smyth/SquadJS/blob/master/plugins/team-randomizer/index.js

Sorry I was too fast there I see...
I now re-checked and don't see any other remaining. But do check it again to be 100% sure.

Also I did check the execution time with a regex, match and .toLowerCase()
They are all same there is not even 0,001 ms diffrence. I just picked the most efficient way to fix this issue. If you want I can make them all regex or match.  But I don't see any reason to do so 😄

LeventHAN:
changed with #82

PR #80, opened 2020-09-24
Better updates

SeanWalsh95:
* ignore config.json

this is to prevent accidently uploading sensitive information that may be in config.json as well as the ability to update SquadJS via git pull without overriding config.json as it is no longer tracked

* replace config.json with example-config.json

replaces the initial function of config.json with example-config.json and adds code to create config.json from example-config.json if it doesn’t already exist on initial run


LeventHAN (changes requested):
Perhaps change the error to :
SquadJS couldn't find any configuration file. We have generated a config file at (#location). The file needs to be configured before run.

For the rest 👍

SeanWalsh95:
@11TStudio Message updated in 2c5aa12

LeventHAN (approved):

Thomas-Smyth (changes requested):
There are a few other things that need doing to make this complete:
 * Update the documentation in the [readme template](https://github.com/Thomas-Smyth/SquadJS/blob/master/factory/templates/readme-template.md) to explain the copying process.
 * Update the [`build-config`](https://github.com/Thomas-Smyth/SquadJS/blob/master/factory/build-config.js) script to write to the file's new name.
 * Build config and readme with `yarn run build-all`.

Whilst I agree this is a good approach to protect credentials, and have used it in the past myself, I feel that it might not be the best approach for SquadJS. Unlike some other projects where this method is used, such as Lavarel, SquadJS's users are less likely to be experienced in coding, git, etc. potentially making this additional step confusing to them and also a little irrelevant to them if they do not intend to commit. Instead, it might be better to suggest in the documentation to the more knowledgable users who intend to commit code to use the `config-test.json` and `index-test.js` files as they are more likely to be able to cope with this additional step and the additional step is more relevant to them.

Alternatively, it might be an even better idea to add a pre-commit check, like the ESLint/Prettier checks, that compares the config with the [template config](https://github.com/Thomas-Smyth/SquadJS/blob/master/factory/templates/config-template.json) to see whether credentials have been placed into the file. This would ultimately be the best approach as it would prevent anyone from committing credentials without any additional steps.

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Project Files
 *.tmp
 
+config.json
 index-test.js
```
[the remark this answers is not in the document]
SeanWalsh95: Will do, my original idea for this was just to make it easier to update, the credentials thing was something I thought of after the fact.

The original idea being that you could use git to distribute updates for the common user that's not going to change anything other than the config so the only thing they would have to do at that point is pull down the changes from master making any update easy to do.

factory/read-config.js
```diff
@@ -6,7 +6,15 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 export default function (configPath = './config.json') {
   const fullConfigPath = path.resolve(__dirname, '../', configPath);
-  if (!fs.existsSync(fullConfigPath)) throw new Error('Config file does not exist.');
+  const fullExamplePath = path.resolve(__dirname, '../', './example-config.json');
+
+  if (!fs.existsSync(fullConfigPath)) {
+    fs.copyFile(fullExamplePath, fullConfigPath, (err) => {
```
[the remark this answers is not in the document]
SeanWalsh95: should be resolved in 17b11b5

.gitignore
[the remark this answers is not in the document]
Thomas-Smyth: I hadn't thought of that when replying originally, but I don't think it changes anything.

Even after adding credentials `git pull` should still work in most cases. The only cases that I can think of where merge conflicts occur are when a significant change has been made to the config or a default value has been changed, which is quite rare and when does occur is well documented on Discord, etc.. The benefit of allowing git to update configs in most cases it will be handled fine and new plugins/options will be added without users having to manually copy them to their config. Having said that, I don't experience updating from a user perspective much, so I not sure whether others feel differently?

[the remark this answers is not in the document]
SeanWalsh95: my understanding is any tracked file will be overwritten via `git pull` by the the file from the repo if it was edited but not committed/stashed, the idea being to not track the config file in order to preserve at least the basic settings for a server.

I'm almost thinking it might be best to split it into two files server_config and plugins_config so that the things that never really change can be in a static untracked file (server_config) and the plugins can be overwritten with the newest recommended default options, I'd think this would work for 90% of users and also provide a command to give to users for a simple way to update (`git pull`).

factory/templates/readme-template.md
```diff
@@ -33,7 +33,7 @@ SquadJS relies on being able to access the Squad server log directory in order t
 ### Installation
 1. Clone the repository: `git clone https://github.com/Thomas-Smyth/SquadJS`
 2. Install the dependencies: `yarn install`
-3. Configure the `config.json` file. See below for more details.
+3. Copy `example-config.json` to `config.json` and configure it. See below for more details.
```
Thomas-Smyth: > Copy `example-config.json` and rename to `config.json` and configure it. See below for more details.

Probably more clear that you need to create the new file.

factory/read-config.js
```diff
@@ -6,7 +6,15 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
...
 export default function (configPath = './config.json') {
   const fullConfigPath = path.resolve(__dirname, '../', configPath);
-  if (!fs.existsSync(fullConfigPath)) throw new Error('Config file does not exist.');
+  const fullExamplePath = path.resolve(__dirname, '../', './example-config.json');
+
+  if (!fs.existsSync(fullConfigPath)) {
+    try{
+      fs.copyFileSync(fullExamplePath, fullConfigPath);
+      throw new Error("SquadJS couldn't find any configuration file. We have generated a config file at "+fullConfigPath+". This file needs to be configured before running JquadJS again.");
+    } catch (err) { throw err; }
```
Thomas-Smyth: Can get rid of the try/catch here as you're just rethrowing the error.

factory/read-config.js
```diff
@@ -6,7 +6,15 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 export default function (configPath = './config.json') {
   const fullConfigPath = path.resolve(__dirname, '../', configPath);
-  if (!fs.existsSync(fullConfigPath)) throw new Error('Config file does not exist.');
+  const fullExamplePath = path.resolve(__dirname, '../', './example-config.json');
+
+  if (!fs.existsSync(fullConfigPath)) {
+    try{
+      fs.copyFileSync(fullExamplePath, fullConfigPath);
+      throw new Error("SquadJS couldn't find any configuration file. We have generated a config file at "+fullConfigPath+". This file needs to be configured before running JquadJS again.");
```
Thomas-Smyth: Use a template literal here please for consistency's stake.
```
throw new Error(`SquadJS couldn't find any configuration file. We have generated a config file at ${fullConfigPath}. This file needs to be configured before running SquadJS again.`);
```

There is also a typo in SquadJS.

Thomas-Smyth (changes requested):

factory/read-config.js
SeanWalsh95: the try catch is just for the case where the original file is missing, but is suppose they have bigger problems if that's the case

Thomas-Smyth: The `if (!fs.existsSync(fullConfigPath)) {` block catches the cases if the file does not exist. There's no point in the try catch block as it's catching the error and just throwing it again.

SeanWalsh95:
I'm reposting my comment because I think it may have been missed.

> my understanding is any tracked file will be overwritten via `git pull` by the the file from the repo if it was edited but not committed/stashed, the idea being to not track the config file in order to preserve at least the basic settings for a server.
> 
> I'm almost thinking it might be best to split it into two files server_config and plugins_config so that the things that never really change can be in a static untracked file (server_config) and the plugins can be overwritten with the newest recommended default options as most people wont change them I'd think this would work for 90% of users and also provide a command to give to users for a simple way to update (`git pull`).

@Thomas-Smyth 
I have written an alternative parser for the config file to do just this what do you think? I wanted to check with you before I get too far ahead of myself on this.

SeanWalsh95:
I believe I have come up with a much better solution than my original.

SeanWalsh95:
Instead of splitting up the file, this untracks the standard config file and merges it with the example-config  on runtime allowing for plugins to be added after the fact and for SquadJS to automatically update the untracked config file with any new plugins/settings preserving the user defined settings.

SeanWalsh95:
closing in favor of merging from branch instead of master

PR #82, opened 2020-09-28
commands made case insensitive

LeventHAN:
Admin and vote commands are now  case insensitive.
(Hopefully this time on the correct base and branch)


Thomas-Smyth:
Thanks for the PR.

Going to implement this slightly differently in v2 so that events can be bound to `CHAT_COMMAND:<command>` where `command` is always lowercase.

https://github.com/Thomas-Smyth/SquadJS/pull/90/files#diff-2e28c560e79acdb77e8d4703945e1ec3R160

PR #86, opened 2020-10-02
Discord server status update timer

lukaslpzi:
A timer that automaticley updates discord server status.
Manual refresh works the same way as before.
New param for the refresh interval in seconds.


Thomas-Smyth:
This solution isn't very persistent as when the bot restarts the refreshing will be lost. A better solution has been implemented in v2 with a possibly better again solution being developed.

PR #89, opened 2020-10-04
Better updates

SeanWalsh95:


SeanWalsh95:
closed in favor of adding to v2

PR #92, opened 2020-10-08
Refactor squad-server getPlayerBy* helpers

bumbummen99:
Makes the squad-server a bit more DRY by abstracting getPlayerBy* helpers into an getPlayerByAttribute helper.


Thomas-Smyth:
Thanks for the PR, but this has already been handled in v2: https://github.com/Thomas-Smyth/SquadJS/pull/90/files#diff-2e28c560e79acdb77e8d4703945e1ec3R252

PR #93, opened 2020-10-09
Add ftpHost option

bumbummen99:
Allow to use ftpHost if set, otherwise fall back to host.
This is a non-breaking change, to use it you will have to add `ftpHost` to your configuration.


Thomas-Smyth:
Added to v2: https://github.com/Thomas-Smyth/SquadJS/pull/90/commits/f923e9179e6d116993dc0ffc2b920066fedbc560

Will leave it hidden from the config as I doubt it will be used much.

Should we do the same for RCON as if I recall correctly that can be set to a different IP too?

PR #97, opened 2020-10-21
Auto Merge Config

SeanWalsh95:
this adds the ability to merge the users config file with an example/template, allowing for an option to update the users config it it is missing any new keys/plugins. This also untracks the users config file from git allowing for the ability to update via `git pull` without squashing the current config with the one currently on github, this also has the nice side effect of helping to not accidentally upload your config if you are contributing to the project, or somehow accidently pushed your config via git.


Thomas-Smyth:
Going to prioritise other stuff for the time being and come back to this on the basis that I think it will only be applicable moving from SquadJS v2.0.0 to v2.0.1+ and I don't want to revisit it each time we change the config style whilst we work on v2.

PR #101, opened 2020-10-24
Rework MapVotes for V2

wojski:
Rewrite and adjust original mapvotes (Mapvote did you mean, Mapvote 123) to work with V2.

Filters are not implemented for "Mapvote did you mean" yet.

In next step I will focus on usage of filters + some rewrite map filter to allow plugin to provide custom filters.


wojski:
Filters for mapvote autocomplete implemented, tested and working

squad-server/plugins/mapvote/components/mapvote-vote-engine.js
```diff
@@ -0,0 +1,138 @@
+import { MAP_VOTE_NEW_WINNER, MAP_VOTE_START, MAP_VOTE_END } from 'mapvote/constants';
+import EventEmitter from 'events';
+
+export default class MapvoteVoteEngine extends EventEmitter {
```
Thomas-Smyth, line 4: May I suggest we use a more OOP approach no we can and move away from the `MapVote` class and event emitter?

For example, when a new map is voted for both options are going to take a layer name, cast the vote count, find the new winner, update the next map, and announce. An example of whether they differ would be in the announcement of a new winner as `mapvote-123` map need to announce the layer numbers. So we should have a method in a base class called, for example, `announceCurrentWinner` that simply announces the winner and is called at the end of the above logic. `mapvote-123` can then overwrite this method to include the layer numbers.

Thomas-Smyth (changes requested):
I think we should rethink the approach given now we use proper classes we can make more use of OOP in particular extending/overwriting. We should have a `MapvoteBasePlugin` that contains the common logic, e.g. calculating a winner, for each version. If we intend to overwrite some of this then we should separate this section of logic out into separate methods we can easily overwrite. For example, as discussed below, the announcement of a winner will differ between version so we should separate this out into a single basic method that can be easily overwritten by each version without needing to repeat any of the common logic.

As a side note, I'd suggest, where possible, making separate smaller PRs for each aspect of these large changes, e.g. make the event constants are separate. That way we can move forwards with the good stuff whilst rethinking the not so good stuff.

Thomas-Smyth:
Closing this as map voting will become pointless with the release of Squad v2. Thanks for the PR.

PR #102, opened 2020-10-25
Auto Kick Unassigned Plugin

SeanWalsh95:


lukaslpzi:
From the rapid glance i can se multiple iprovement points:
- It should not be called "Auto Kick AFK", its rather "Auto Kick Unasigned"
- Admin list should be a connector (like map layers)
- Admin identification method should take admin group in parameter, most likley array of strings
- Auto kick, and auto warn should be separated plugins, you cant use autowarn alone.
- Kick warning could have countdown (ex 4/6 warnings, last warning etc)

SeanWalsh95:
* As it does also kick unassigned players the warnings are supposed to mitigate that as much as possible as the intent is to automatically kick AFK players.
* will do
* unsure what you mean, is the suggestion to store admin groups and admins separately instead of merging them into one admin object?
* I could possibly add the feature of just warning players if its wanted although it would be simple to do now as a separate plugin
* This already kind of exists in the form form of a timer at the end of the warning message it is a countdown to a kick and is generated based off the set options.

lukaslpzi:
1. Your intent may be to create such plugin, et as for now the code selects  only unasigned players for watch, they may be afk, they may not. Same for afk players in squads, this code does not watch over them.
2. nice
3. No, i sugest that every server may have diferent way of seperating admins. I would rather see whitelise of groups from remote admin list that will not get kicked. This one would be a plugin option.
4. just a loose sugestion.
5. im blind.

SeanWalsh95:
1. Yeah I suppose you're right ill get to renaming things, hopefully SL's will be vigilant enough to kick players in their squad that aren't playing but this also doesn't account for the case where there is an entire squad afk including the SL, or just an afk SL .
2. Still trying to wrap my head around how to implement this, the first spot was just because it worked
3. Do you think it would be beneficial to have `server.admins`, `server.whitelist`, `server.adminGroups` lists where the first two are the steamID/groupID pairs and adminGroups is a set of groupID/perms?
4. I'll keep it in mind.

lukaslpzi:
1. There was and idea to determinate if plaers are afk based on the points they gain, still i think its a huge bodge and i dont know any way of finding afk players.
3. It could be something of this kind, but i doubt there is any use in deviding into admin and whitelist. Someone who configures this plugin should know how the server is configured and what are the names of groups that he wants to exclude from being kicked.

SeanWalsh95:
@lukaslpzi your suggestions should now be implemented

rcon/index.js
```diff
@@ -362,4 +362,16 @@ export default class Rcon extends EventEmitter {
...
+
+  async warn(steamID, message) {
+    await this.execute(`AdminWarn "${steamID}" ${message}`);
+  }
+
+  async kick(steamID, reason) {
+    await this.execute(`AdminKick "${steamID}" ${reason}`);
+  }
+
+  async switchTeam(steamID) {
```
Thomas-Smyth: I'd keep these method names similar to the commands, i.e. `forceTeamChange`, on the basis this is what most Squad admins will know the commands as.

squad-server/index.js
```diff
@@ -252,6 +252,48 @@ export default class SquadServer extends EventEmitter {
...
+      const remoteAdmins = {
+        admins: {},
+        whitelist: {},
+        groups: {}
+      };
+
+      for (let idx = 0; idx < remoteAdminLists.length; idx++) {
+        const list = remoteAdminLists[idx];
+
+        const resp = await axios({
```
Thomas-Smyth: I'd suggest moving this logic elsewhere on the basis that may want to support other methods of getting admin lists too, e.g. locally and perhaps over FTP, so this method might get quite long. Create a file in `utils` that exports a function called `fetchAdminLists` that takes the config, as shown below, that is then called within this method with the resulting being updated in the class' state. I'd also consider adding this like the `updatePlayerList`, etc. methods so it is updated over time (just less frequently or maybe even just after a map change).

Each admin list probably needs a type and details of the source, e.g.:
```
{ "type": "remote", "url": "HTTP://...." }
```
or 
```
{ "type": "local", "url": "path/to/file" }
```

squad-server/index.js
```diff
@@ -252,6 +252,48 @@ export default class SquadServer extends EventEmitter {
...
+        const adminRgx = /(?<=Admin=)(\d+):(\S+)(?=\s)/g;
+
+        /* eslint-disable no-unused-vars */
+        for (const [match, groupID, groupPerms] of rawData.matchAll(groupRgx)) {
+          remoteAdmins.groups[`${idx}-${groupID}`] = groupPerms.split(',');
+        }
+        for (const [match, steamID, groupID] of rawData.matchAll(adminRgx)) {
+          const perms = remoteAdmins.groups[`${idx}-${groupID}`];
+          if (!(perms.includes('reserve') && perms.length === 1)) {
+            remoteAdmins.admins[steamID] = `${idx}-${groupID}`;
```
Thomas-Smyth: I'd suggest just a single list/set to store the Steam ID with the perms, e.g.:
```
[
  {
    steamID: "xxx",
    permA: true.
    permB: true
    // etc.
  }
]
```

Then add methods to filter this, e.g. `getAdminInfoBySteamID(steamID)` or `getAdminsWithPermission(permA)`, similar to the player list.

squad-server/index.js
```diff
@@ -252,6 +252,48 @@ export default class SquadServer extends EventEmitter {
...
+          } else {
+            remoteAdmins.whitelist[steamID] = `${idx}-${groupID}`;
+          }
+        }
+        /* eslint-enable no-unused-vars */
+      }
+      Logger.verbose('SquadServer', 3, 'RemoteAdmins:', remoteAdmins);
```
Thomas-Smyth: I think we should try to keep logs to a single line if they're not an error. Might be better to log each admin as they're processed in the loop and then perhaps `x admins loaded` at the end.

squad-server/index.js
```diff
@@ -265,6 +307,14 @@ export default class SquadServer extends EventEmitter {
         ...oldPlayerInfo[player.steamID],
         ...player
       }));
+
+      for (const player of this.players) {
+        if (typeof oldPlayerInfo[player.steamID] === 'undefined') continue;
+        if (player.teamID !== oldPlayerInfo[player.steamID].teamID)
+          this.emit('PLAYER_TEAM_CHANGE', player);
+        if (player.squadID !== oldPlayerInfo[player.steamID].squadID)
+          this.emit('PLAYER_SQUAD_CHANGE', player);
```
Thomas-Smyth: Maybe we should pass their old team/squad, including if that was undefined. Perhaps in the format:
```
{
  player: { ... },
  old: x, // can be undefined if they just joined.
  new: y
}
```

squad-server/index.js
```diff
@@ -456,6 +509,9 @@ export default class SquadServer extends EventEmitter {
             connectorConfig.filter,
             connectorConfig.activeLayerFilter
           );
+        } else if (option.connector === 'remoteAdminLists') {
+          Logger.verbose('SquadServer', 1, `Starting remoteAdminList connector...`);
+          connectors[connectorName] = await server.setupAdminList(connectorConfig);
```
Thomas-Smyth: The admin list is information about the server so can go into the `SquadServer` class rather than being a connector. Connectors should only be used for stuff that's not directly related to the server, e.g. Discord, MySQL or subsets of Squad maps (i.e. rotations/pools).

squad-server/plugins/auto-kick-unassigned.js
```diff
@@ -0,0 +1,225 @@
...
+        description: 'How long in minutes to wait before a player that is unassigned is kicked',
+        default: 6
+      },
+      playerThreshold: {
+        required: false,
+        description:
+          'Player count required for AutoKick to start kicking players to disable set to -1 to disable',
+        default: 93
+      },
+      queueThreshold: {
```
Thomas-Smyth: Is this needed or just added complexity? If the server is full you probably should kick AFKs so there is no queue and people can join anyway. Using the player threshold to keep the server between 90-100 would make it look full, but not be entirely full.

squad-server/plugins/auto-kick-unassigned.js
```diff
@@ -0,0 +1,225 @@
...
+   *      startTime: <Epoch Date>
+   *    warnTimerID: <intervalID>
+   *    kickTimerID: <timeoutID>
+   *  }
+   */
+  constructor(server, options) {
+    super();
+
+    this.server = server;
+    this.options = options;
```
Thomas-Smyth: If you extend `BasePlugin`, just do `super(server, options, rawOptions)` and scrap the last two lines.

squad-server/plugins/auto-kick-unassigned.js
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    if (forceUpdate) await this.server.updatePlayerList();
+
+    // loop through players on server and start tracking players not in a squad
+    for (const player of this.server.players) {
+      const isTracked     = player.steamID in this.trackedPlayers;
+      const isUnassigned  = player.squadID === null;
+      const isAdmin       = player.steamID in this.options.adminList.admins;
+      const isWhitelist   = player.steamID in this.options.adminList.whitelist;
```
Thomas-Smyth: As discussed above, we should manage this info this inside the `SquadServer` class. So maybe add a method `hasAdminPerm(perm, steamID)` to return true/false and then continue in the loop as appropriate.

squad-server/plugins/auto-kick-unassigned.js
```diff
@@ -0,0 +1,225 @@
...
+
+      if (isUnassigned && isAdmin) Logger.verbose('AutoKick', 2, `Admin is Unassigned: ${player.name}`);
+      if (isUnassigned && isWhitelist) Logger.verbose('AutoKick', 2, `Whitelist player is Unassigned: ${player.name}`);
+
+      // start tracking player
+      if (isUnassigned && !isTracked && !(isAdmin && this.options.ignoreAdmins) && !(isWhitelist && this.options.ignoreWhitelist))
+        this.trackedPlayers[player.steamID] = this.trackPlayer(player);
+
+      // tracked player joined a squad remove them (redundant afer adding PLAYER_SQUAD_CHANGE, keeping for now)
+      if (!isUnassigned && isTracked) this.untrackPlayer(player.steamID);
```
Thomas-Smyth: It's better to reorder conditions so they're less repetitive, e.g. do this line first and then `continue` to skip to the next player in the loop to avoid needing to do `if(isUnassigned` 3 times.

Thomas-Smyth (changes requested):
Looks ok... Will test fully in-game once these minor changes have been made.

squad-server/index.js
SeanWalsh95: I have done essentially this but instead of a list it is a dictionary for easy lookups of steamID's

squad-server/index.js
SeanWalsh95: I believe I have done this the way you asked

squad-server/plugins/auto-kick-unassigned.js
SeanWalsh95: yeah, probably is I was just thinking of things you could do and never stopped to think if I should :). I've removed it.

Thomas-Smyth (approved):

PR #106, opened 2020-12-08
Fixing NaNms

LeventHAN:


Thomas-Smyth:
The `Nan`s actually make more sense here than `0`. `0` implies that logs are being parsed with a 0ms latency when in reality it's a matter of no logs being parsed at all. I don't think there's any harm to having `NaN` in here.

11TStudio:
just write in place of the zero "No Logs are parsed" 😂

PR #107, opened 2020-12-08
SquadJS logging for MySQL

SeanWalsh95:
Sends Sequelize logging to Logger.verbose so that it is toggleable via SquadJS configs

### DIFF
--- a/squad-server/factory.js
+++ b/squad-server/factory.js
@@ -114,7 +114,7 @@ export default class SquadServerFactory {
     }
 
     if (type === 'sequelize') {
-      let mergedConfig = Object.assign(connectorConfig, {logging: msg => Logger.verbose('MySQL', 1, msg)} );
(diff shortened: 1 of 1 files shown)


squad-server/factory.js
```diff
@@ -114,7 +114,8 @@ export default class SquadServerFactory {
     }
 
     if (type === 'sequelize') {
-      const connector = new Sequelize(connectorConfig);
+      let mergedConfig = Object.assign(connectorConfig, {logging: msg => Logger.verbose('MySQL', 1, msg)} );
+      const connector = new Sequelize(mergedConfig);
```
Thomas-Smyth, line 118: Good spot, but might be better as:

```js
const connector = new Sequelize({ ...connectorConfig, logging: msg => Logger.verbose('Sequelize', 3, msg) });
```

PR #113, opened 2020-12-22
Edits to auto-kick-unassigned docs

SeanWalsh95:


squad-server/plugins/auto-kick-unassigned.js
```diff
@@ -26,35 +26,43 @@ export default class AutoKickUnassigned extends BasePlugin {
...
         required: false,
-        description: 'How often in seconds should we warn the player about being unassigned?',
+        description:
+          'How often in <b>Seconds</b> should we warn the player about being unassigned?',
         default: 30
       },
       unassignedTimer: {
         required: false,
-        description: 'How long in minutes to wait before a player that is unassigned is kicked',
+        description: 'How long in <b>Minutes</b> to wait before a unassigned player is kicked',
```
Thomas-Smyth: I think it would be best to keep the config using a consistent time unit, i.e. all in seconds. That would save this kind of confusion.

Thomas-Smyth (approved):

PR #119, opened 2021-01-07
updated config and readme for adminlists

werewolfboy13:


got2bhockey:
@werewolfboy13 You'll need to update this file as well to have the readme not get overwritten when it is regenerated:
https://github.com/Thomas-Smyth/SquadJS/blob/beta/squad-server/templates/readme-template.md

Thomas-Smyth:
I think it would be a good idea to add the full examples removed in this commit: https://github.com/Thomas-Smyth/SquadJS/commit/23518b4d976a89ae96dc4d32098341560998789d

That way people can see examples of how to use `remote` and `local` as the options.

werewolfboy13:
How do you think on this commits? Want it worded any way different?

Thomas-Smyth:
I would havve just added it to the example above that, exactly how it was before it was removed.

werewolfboy13:
This better?

Thomas-Smyth:
Yeah, that will do, thanks.

werewolfboy13:
Thanks I forgot to undo the indentation that VS likes to do.

Issue #122, opened 2021-01-10
Fatal Crash on admins list update with tk bot


werewolfboy13:
### Description of Issue

With the tk bot (im thinking this is the issue) a fatal crash occurs.

### Errors or Screenshots of Issue

TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at SquadServer.getAdminsWithPermission (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/squad-server/index.js:248:43)
    at new AutoKickUnassigned (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/squad-server/plugins/auto-kick-unassigned.js:86:26)
    at Function.buildFromConfig (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/squad-server/factory.js:89:22)
    at processTicksAndRejections (node:internal/process/task_queues:93:5)
    at async main (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/index.js:14:7)
file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/squad-server/index.js:248
    for (const [steamID, perms] of Object.entries(this.admins)) {
                                          ^

TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at SquadServer.getAdminsWithPermission (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/squad-server/index.js:248:43)
    at new AutoKickUnassigned (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/squad-server/plugins/auto-kick-unassigned.js:86:26)
    at Function.buildFromConfig (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/squad-server/factory.js:89:22)
    at processTicksAndRejections (node:internal/process/task_queues:93:5)
    at async main (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/index.js:14:7)

### Squad Information
 * SquadJS Version: current beta branch
 * Squad Version: current

### System Information
 * Operating System: 2016 and 2019 windows server
 * NodeJS Version: current
 * Yarn Version: current

werewolfboy13:
![image](https://user-images.githubusercontent.com/43587009/104112434-f8988e00-5318-11eb-9ff0-3a94c3d54f66.png)

[...]

Issue #124, opened 2021-01-10
AutoKickUnassigned is still kicking admins


werewolfboy13:
### Description of Issue

The auto AutoKickUnassigned is not ignoring admins.

### Errors or Screenshots of Issue

Config:

      "plugin": "AutoKickUnassigned",
      "enabled": true,
      "warningMessage": "Join a squad, you are are unassigned and will be kicked",
      "kickMessage": "Unassigned - automatically removed",
      "frequencyOfWarnings": 60,
      "unassignedTimer": 300,
      "playerThreshold": 93,
      "roundStartDelay": 900,
      "ignoreAdmins": true,
      "ignoreWhitelist": false

`    // Custom Permissions here for admins.`
`    this.admins = server.getAdminsWithPermission('forceteamchange');`
`    this.whitelist = server.getAdminsWithPermission('reserve');`

I had to change the custom permission. Nothing I change keeps the admins from being kicked.

### Squad Information
 * Plugins in Use: 
 * SquadJS Version: beta brance
 * Squad Version: current

### System Information
 * Operating System: windows server 2019
 * NodeJS Version: current
 * Yarn Version: current

werewolfboy13:
Issue doesn't appear to be fixed.

SeanWalsh95:
set `AutoKickUnassigned` verbose to 2, is it even identifying the admins?

werewolfboy13:
![image](https://user-images.githubusercontent.com/43587009/104379067-223cf980-5553-11eb-9f8a-7437e678c82e.png)
this is a reload. all it shows

werewolfboy13:
![image](https://user-images.githubusercontent.com/43587009/104379115-32ed6f80-5553-11eb-88f3-1cdfd505f9a4.png)

werewolfboy13:
![image](https://user-images.githubusercontent.com/43587009/104379282-78aa3800-5553-11eb-85cd-9f3fd3c3d476.png)
![image](https://user-images.githubusercontent.com/43587009/104379302-7e078280-5553-11eb-8aaf-5529b6587684.png)

SeanWalsh95:
seems to be pulling the admins, looking to see if the plugin is actually logging admins something like `Admin is Unassigned: <playername>`

werewolfboy13:
waiting for the kick now. Also on another note the bot is still warning after a player joins a squad.

SeanWalsh95:
[...]

PR #125, opened 2021-01-12
Auto-Kick-Unassigned patch

SeanWalsh95:
patch to resolve issue #122, moves getAdminList functions to plugin mount so the SquadServer instance is created and pulls the lists before the plugin attempts to get them.


SeanWalsh95:
5c5be92 is in regards to Issue #124

Thomas-Smyth:
Although this might fix the issue, I do not think it's the best fix.

The error in the issue is caused by the `admins` property in the `SquadServer` class is not defined until after the admin list is fetched. This means when the plugins calls the `getAdminsWithPermission` method the `Object.entries` fails on the undefined property. To fix this, we can simply define `admins` as an empty object in the constructor of `SquadServer` so `Object.entries` does not throw the error.

However, the population of this `admins` property is now also called later which causes the `getAdminsWithPermission` in the plugin's constructor to return no data even with the above in place. We should get around this by not storing a copy of this data inside the plugin and by referring to `getAdminsWithPermission` each time the data is needed. This is probably a better idea anyway as we probably should make the admin lists update periodically in the future to avoid the need for SquadJS to be restarted for changes to the list to have an effect.  Storing a copy of the data inside the plugin will make it harder to update.

As a side note, neither https://github.com/Thomas-Smyth/SquadJS/blob/beta/squad-server/index.js#L242 nor https://github.com/Thomas-Smyth/SquadJS/blob/beta/squad-server/index.js#L246 need to be asynchronous methods as no asynchronous actions are made within them. We should, therefore, remove the `async` keyword from the method and any `await` keywords from where it is called.

PR #130, opened 2021-01-14
Updated Year For Copyright Message

werewolfboy13:


squad-server/utils/constants.js
```diff
@@ -9,6 +9,6 @@ const SQUADJS_VERSION = JSON.parse(
 ).version;
 
 /* As set out by the terms of the license, the following should not be modified. */
-const COPYRIGHT_MESSAGE = 'Powered by SquadJS, Copyright © 2020';
+const COPYRIGHT_MESSAGE = 'Powered by SquadJS, Copyright © 2021';
```
SeanWalsh95: It might be better to do something like the following:
``` `Powered by SquadJS, Copyright © ${new Date().getFullYear()}` ```

werewolfboy13: Is the update better?

SeanWalsh95: 👍

PR #136, opened 2021-01-30
Update index.js

LeventHAN:
adding a condition before it trys to put something in an object. Since on the first fresh start of the server it try to put null and this throws errors at the logs...
Dunno if someone else is getting an error but this fixed my error @ the terminal


Thomas-Smyth:
`this.admins` should always be defined. If it's not then we need to fix that error and not do the above,

LeventHAN:
The issue was only on my side... After a fresh clone it worked just fine. I guess I did something wrong when fetching >.<

PR #141, opened 2021-02-25
set pulled flag + logging

SeanWalsh95:
Wraps axios request with try/catch for cases where we do not get a response or get an error

sets `this.pulled` flag after receiving and parsing response


Thomas-Smyth:
If failure to fetch layers is hidden then this is going to cause issues down the line, e.g. when SquadJS tries to get info on a layer?

The changes to the pulled flag make sense though.

PR #142, opened 2021-02-25
Intervalled Broadcasts plugin causes a crash

milutinke:
Line 47 causes a crash because of undefined. The code should have been `this.options.broadcasts` instead of `this.broadcasts`.
This would happen when the plugin tried to broadcast a message.


Thomas-Smyth:
Good spot, thanks.

Issue #148, opened 2021-02-27
If a player name has a special character an error throws in mySQL plugin.


werewolfboy13:
### Description of Issue
If a player name has a special character like 😲 an error throws.
### Errors or Screenshots of Issue

[SquadServer][1] Updating player list...
DatabaseError [SequelizeDatabaseError]: Conversion from collation utf8mb4_unicode_ci into utf8_general_ci impossible for parameter
    at Query.formatError (C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\sequelize\lib\dialects\mysql\query.js:239:16)
    at Query.run (C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\sequelize\lib\dialects\mysql\query.js:54:18)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\sequelize\lib\sequelize.js:619:16
    at async MySQLQueryInterface.upsert (C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\sequelize\lib\dialects\mysql\query-interface.js:52:12)
    at async Function.upsert (C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\sequelize\lib\model.js:2482:20)
    at async DBLog.onPlayerWounded (file:///C:/Users/Administrator/Desktop/Servers/Spaghetti-Bot/squad-server/plugins/db-log.js:459:7) {
  parent: Error: Conversion from collation utf8mb4_unicode_ci into utf8_general_ci impossible for parameter
      at Packet.asError (C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\mysql2\lib\packets\packet.js:712:17)
      at Execute.execute (C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\mysql2\lib\commands\command.js:28:26)
      at Connection.handlePacket (C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\mysql2\lib\connection.js:425:32)
      at PacketParser.onPacket (C:\Users\Administrator\Desktop\Servers\Spaghetti-Bot\node_modules\mysql2\lib\connection.js:75:12)
[...]

PR #154, opened 2021-03-01
Fixed bug #148. Added PLAYER_DISCONNECTED event

milutinke:
Sorry for pushing to master, I could not merge it with my copy of the beta branch for some reason, I am not that experienced with git.
#148 is fixed.
Added PLAYER_DISCONNECTED event based on EAC UnregisterClient event.
Also, I did not want to mess anything by installing the packages with yarn and then linting and deleting the packages, so it should be linted.

### DIFF
--- a/squad-server/index.js
+++ b/squad-server/index.js
@@ -178,6 +178,16 @@ export default class SquadServer extends EventEmitter {
       this.emit('PLAYER_CONNECTED', data);
     });
 
+    this.logParser.on('PLAYER_DISCONNECTED', async (data) => {
+      data.player = await this.getPlayerBySteamID(data.steamID);
+      if (data.player) data.player.suffix = data.playerSuffix;
+
+      delete data.steamID;
+      delete data.playerSuffix;
+
+      this.emit('PLAYER_DISCONNECTED', data);
+    });
+
     this.logParser.on('PLAYER_DAMAGED', async (data) => {
       data.victim = await this.getPlayerByName(data.victimName);
       data.attacker = await this.getPlayerByName(data.attackerName);
--- a/squad-server/log-parser/player-disconnected.js
+++ b/squad-server/log-parser/player-disconnected.js
@@ -1,0 +1,19 @@
+export default {
+    regex: /^\[([0-9.:-]+)]\[([ 0-9]*)]LogEasyAntiCheatServer: \[[0-9:]+]\[[A-z]+]\[EAC Server] \[Info]\[UnregisterClient] Client: ([A-z0-9]+) PlayerGUID: ([0-9]{17})/,
+    onMatch: (args, logParser) => {
+        const playerTrackerObjectName = `player-name-tracker-${args[4]}`;
+
+        const data = {
+            raw: args[0],
+            time: args[1],
+            chainID: args[2],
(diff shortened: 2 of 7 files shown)


squad-server/log-parser/player-disconnected.js
```diff
@@ -0,0 +1,19 @@
+export default {
+    regex: /^\[([0-9.:-]+)]\[([ 0-9]*)]LogEasyAntiCheatServer: \[[0-9:]+]\[[A-z]+]\[EAC Server] \[Info]\[UnregisterClient] Client: ([A-z0-9]+) PlayerGUID: ([0-9]{17})/,
+    onMatch: (args, logParser) => {
+        const playerTrackerObjectName = `player-name-tracker-${args[4]}`;
+
+        const data = {
+            raw: args[0],
+            time: args[1],
+            chainID: args[2],
+            playerSuffix: logParser.eventStore[playerTrackerObjectName] ? logParser.eventStore[playerTrackerObjectName] : 'Unknown',
```
Thomas-Smyth: There's probably no need to be working with the player's suffix here (or anywhere in this PR).

The player's suffix is collected using two log events and then is added into their player object listed in the `SquadServer.players` array. As this log line has their SteamID in the `SquadServer` class, where you forward the event, you can use the `.getPlayerBySteamID()` method to get the player's object. Then you can probably remove all other references to the player's suffix from this PR.

Thomas-Smyth (changes requested):
Other than the comment regarding the player's suffix this PR looks good to me.

squad-server/log-parser/player-disconnected.js
milutinke: Oh, okay, I did not know it. I will remove that part.

squad-server/index.js
```diff
@@ -178,6 +178,16 @@ export default class SquadServer extends EventEmitter {
       this.emit('PLAYER_CONNECTED', data);
     });
 
+    this.logParser.on('PLAYER_DISCONNECTED', async (data) => {
+      data.player = await this.getPlayerBySteamID(data.steamID);
+      if (data.player) data.player.suffix = data.playerSuffix;
+
+      delete data.steamID;
+      delete data.playerSuffix;
```
Thomas-Smyth, line 186: Looks like you missed these suffix bits.

Thomas-Smyth (changes requested):

squad-server/log-parser/player-disconnected.js
milutinke: Alright, done. I've removed the reference to the player suffix.
And I am already using `.getPlayerBySteamID()` in `squad-server/index.js`.

squad-server/index.js
```diff
@@ -178,6 +178,16 @@ export default class SquadServer extends EventEmitter {
...
+
+      this.emit('PLAYER_DISCONNECTED', data);
```
SeanWalsh95, line 188: Would this not always emit nothing as we just deleted the info?

milutinke:
Everything should be alright now. 
Can you merge it with the beta branch or I need to do something more?

Thomas-Smyth:
> Everything should be alright now.
> Can you merge it with the beta branch or I need to do something more?

I don't think you've addressed this yet, or at least not pushed it:
https://github.com/Thomas-Smyth/SquadJS/pull/154#discussion_r585848943

PR #161, opened 2021-03-05
show active admins for !admin and warn those + simplify placeholder plugin

LeventHAN:
- Simplify the placeholder plugin I did push before
- Add `warnInGameAdmins` for the `discord-admin-request` plugin.
  - Checks if there are online admins in-game. If there are any online in-game admins the plugins warns the admins with the name of the player and the message. Meanwhile respondig to the player with a message including the amount of active admins.(See screenshot) If there are no active admins, it will show a message saying there are noone but that admins were warned on discord. This will still always do it's default function, warn the admins on discord. But did add just this so in-game admins life is going to be easier by not always switching screens to check if someone did use the command. This will also make the trigger function from BM useless since it is way more effective than a notifaction on your phone 👍  (BM triggers are not free, this is.)
  
    - Please note the above warning is in Turkish, because I did test this on a Turkish server. It is basicly the admin testing the !admin function. This is also why you see 2 warnings, one of them is supposed to go to the admin (the top warning) and the other one to the player requesting it (the bottom warning).
    image

This will fix : https://github.com/Thomas-Smyth/SquadJS/issues/57  and https://github.com/Thomas-Smyth/SquadJS/issues/63

### DIFF
--- a/squad-server/plugins/discord-admin-request.js
+++ b/squad-server/plugins/discord-admin-request.js
@@ -59,14 +59,17 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
         description:
           'Should in-game admins be warned after a players uses the command and should we tell how much admins are active in-game right now.',
         default: false
+      },
+      showInGameAdmins: {
+        required: false,
+        description: 'Should players know how much in-game admins there are active/online?',
+        default: true
       }
     };
   }
 
   constructor(server, options, connectors) {
     super(server, options, connectors);
-
-    this.amountAdmins = 0;
 
     this.lastPing = Date.now() - this.options.pingDelay;
 
@@ -131,39 +134,37 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
 
     await this.sendDiscordMessage(message);
 
-    await this.server.updatePlayerList;
-    let admins = await this.server.getAdminsWithPermission('canseeadminchat');
-
+    const admins = await this.server.getAdminsWithPermission('canseeadminchat');
+    let amountAdmins = 0;
     if (this.options.warnInGameAdmins) {
       for (const player of this.server.players) {
-        if (admins.includes(player.steamID)) {
-          this.amountAdmins++;
-
-          await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
-        }
+        if (!admins.includes(player.steamID)) continue;
+        amountAdmins++;
+        await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
       }
 
-      if (this.amountAdmins === 0) {
+      if (amountAdmins === 0)
         await this.server.rcon.warn(
           info.player.steamID,
           `There are no in-game admins, however an admin has been notified via discord, please wait for us to get back to you.`
         );
-      } else {
+      else if (this.options.showInGameAdmins)
         await this.server.rcon.warn(
           info.player.steamID,
-          `There ${this.amountAdmins > 1 ? 'are' : 'is'} ${this.amountAdmins} in-game admin${
-            this.amountAdmins > 1 ? 's' : ''
+          `There ${amountAdmins > 1 ? 'are' : 'is'} ${amountAdmins} in-game admin${
+            amountAdmins > 1 ? 's' : ''
           }. Please wait for us to get back to you.`
         );
-      }
+      else
+        await this.server.rcon.warn(
+          info.player.steamID,
+          `An admin has been notified, please wait for us to get back to you.`
+        );
     } else {
       await this.server.rcon.warn(
         info.player.steamID,
         `An admin has been notified, please wait for us to get back to you.`
       );
     }
-
-    admins = [];
-    this.amountAdmins = 0;
   }
 }
--- a/squad-server/plugins/discord-admin-request.js
+++ b/squad-server/plugins/discord-admin-request.js
@@ -136,12 +136,11 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
 
     const admins = await this.server.getAdminsWithPermission('canseeadminchat');
     let amountAdmins = 0;
-    if (this.options.warnInGameAdmins) {
-      for (const player of this.server.players) {
-        if (!admins.includes(player.steamID)) continue;
-        amountAdmins++;
+    for (const player of this.server.players) {
+      if (!admins.includes(player.steamID)) continue;
+      amountAdmins++;
+      if (this.options.warnInGameAdmins)
         await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
-      }
     }
 
     if (amountAdmins === 0 && this.options.showInGameAdmins)
--- a/squad-server/plugins/discord-admin-request.js
+++ b/squad-server/plugins/discord-admin-request.js
@@ -142,29 +142,24 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
         amountAdmins++;
         await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
       }
+    }
 
-      if (amountAdmins === 0 && this.options.showInGameAdmins)
-        await this.server.rcon.warn(
-          info.player.steamID,
-          `There are no in-game admins, however an admin has been notified via discord, please wait for us to get back to you.`
-        );
-      else if (this.options.showInGameAdmins)
-        await this.server.rcon.warn(
-          info.player.steamID,
-          `There ${amountAdmins > 1 ? 'are' : 'is'} ${amountAdmins} in-game admin${
-            amountAdmins > 1 ? 's' : ''
-          }. Please wait for us to get back to you.`
-        );
-      else
-        await this.server.rcon.warn(
-          info.player.steamID,
-          `An admin has been notified, please wait for us to get back to you.`
-        );
-    } else {
+    if (amountAdmins === 0 && this.options.showInGameAdmins)
       await this.server.rcon.warn(
         info.player.steamID,
-        `An admin has been notified, please wait for us to get back to you.`
+        `There are no in-game admins, however, an admin has been notified via Discord. Please wait for us to get back to you.`
       );
-    }
+    else if (this.options.showInGameAdmins)
+      await this.server.rcon.warn(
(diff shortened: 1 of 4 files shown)


squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +131,39 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+          `There are no in-game admins, however an admin has been notified via discord, please wait for us to get back to you.`
+        );
+      } else {
+        await this.server.rcon.warn(
+          info.player.steamID,
+          `There ${this.amountAdmins > 1 ? 'are' : 'is'} ${this.amountAdmins} in-game admin${
+            this.amountAdmins > 1 ? 's' : ''
+          }. Please wait for us to get back to you.`
+        );
+      }
```
Thomas-Smyth: I think it would be a good idea to split this into a separate option other than `warnInGameAdmins`. Some may want the warnings for in-game admins but without this because it makes it easier for someone to check whether there's a lack of admins and therefore whether it's a good time to troll.

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +131,39 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
 
     await this.sendDiscordMessage(message);
 
-    await this.server.rcon.warn(
-      info.player.steamID,
-      `An admin has been notified, please wait for us to get back to you.`
-    );
+    await this.server.updatePlayerList;
+    let admins = await this.server.getAdminsWithPermission('canseeadminchat');
```
Thomas-Smyth: You can use `const` here.

squad-server/plugins/discord-admin-request.js
```diff
@@ -53,13 +53,21 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
         required: false,
         description: 'The color of the embed.',
         default: 16761867
+      },
+      warnInGameAdmins: {
+        required: false,
+        description:
+          'Should in-game admins be warned after a players uses the command and should we tell how much admins are active in-game right now.',
```
Thomas-Smyth, line 60: We should specify the permission required, i.e. the admin chat.

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +131,39 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
 
     await this.sendDiscordMessage(message);
 
-    await this.server.rcon.warn(
-      info.player.steamID,
-      `An admin has been notified, please wait for us to get back to you.`
-    );
+    await this.server.updatePlayerList;
```
Thomas-Smyth: This is a method so should have `()` at the end. Either way, not sure it's actually required here?

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +131,39 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+    let admins = await this.server.getAdminsWithPermission('canseeadminchat');
+
+    if (this.options.warnInGameAdmins) {
+      for (const player of this.server.players) {
+        if (admins.includes(player.steamID)) {
```
Thomas-Smyth: For cleaner code, it might be better to negate this and then use `continue`.

```js
for (const player of this.server.players) {
  if (!admins.includes(player.steamID)) continue;

  this.amountAdmins++;
  await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
}
````

File: squad-server/plugins/discord-admin-request.js [hunk 6]
```diff
@@ -123,9 +131,39 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+          this.amountAdmins++;
+
+          await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
+        }
+      }
+
+      if (this.amountAdmins === 0) {
```
Thomas-Smyth: Again, for cleaner code, as this is a single line inside the if block you can use:
```js
if (this.amountAdmins === 0) await ...
```

The same goes for the else block:
```js
else await ...
```

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +131,39 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
-      info.player.steamID,
-      `An admin has been notified, please wait for us to get back to you.`
-    );
+    await this.server.updatePlayerList;
+    let admins = await this.server.getAdminsWithPermission('canseeadminchat');
+
+    if (this.options.warnInGameAdmins) {
+      for (const player of this.server.players) {
+        if (admins.includes(player.steamID)) {
+          this.amountAdmins++;
```
Thomas-Smyth: This can be local to the function. Add before this loop `let adminAmount = 0;` and then replace this with `adminAmount++;`.

You can then remove the `this.adminAmount = 0;` below.

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +131,39 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+      }
+    } else {
+      await this.server.rcon.warn(
+        info.player.steamID,
+        `An admin has been notified, please wait for us to get back to you.`
+      );
+    }
+
+    admins = [];
+    this.amountAdmins = 0;
```
Thomas-Smyth: These two lines aren't needed if you make the changes suggested above.

Thomas-Smyth (changes requested):
Looks good other than the two additions being added under a single option and the nit-picky code style comments.

squad-server/plugins/discord-admin-request.js
LeventHAN: image

squad-server/plugins/discord-admin-request.js
LeventHAN: image

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +134,37 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+          info.player.steamID,
+          `There ${amountAdmins > 1 ? 'are' : 'is'} ${amountAdmins} in-game admin${
+            amountAdmins > 1 ? 's' : ''
+          }. Please wait for us to get back to you.`
+        );
+      else
+        await this.server.rcon.warn(
+          info.player.steamID,
+          `An admin has been notified, please wait for us to get back to you.`
+        );
```
Thomas-Smyth: Thinking about it, should this be dependent on the `warnInGameAdmins` option? Id's suggest making it so you can have this enabled without `warnInGameAdmins` being enabled.

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +134,37 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+      for (const player of this.server.players) {
+        if (!admins.includes(player.steamID)) continue;
+        amountAdmins++;
+        await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
+      }
+
+      if (amountAdmins === 0 && this.options.showInGameAdmins)
+        await this.server.rcon.warn(
+          info.player.steamID,
+          `There are no in-game admins, however an admin has been notified via discord, please wait for us to get back to you.`
```
Thomas-Smyth: `There are no in-game admins, however, an admin has been notified via Discord. Please wait for us to get back to you.`

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +134,37 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+        await this.server.rcon.warn(
+          info.player.steamID,
+          `There ${amountAdmins > 1 ? 'are' : 'is'} ${amountAdmins} in-game admin${
+            amountAdmins > 1 ? 's' : ''
+          }. Please wait for us to get back to you.`
+        );
+      else
+        await this.server.rcon.warn(
+          info.player.steamID,
+          `An admin has been notified, please wait for us to get back to you.`
```
Thomas-Smyth: `An admin has been notified. Please wait for us to get back to you.`

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +134,37 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+        );
+    } else {
+      await this.server.rcon.warn(
+        info.player.steamID,
+        `An admin has been notified, please wait for us to get back to you.`
```
Thomas-Smyth: `An admin has been notified. Please wait for us to get back to you.`

Thomas-Smyth (changes requested):

squad-server/plugins/discord-admin-request.js
LeventHAN: All done 👍

squad-server/plugins/discord-admin-request.js
```diff
@@ -123,9 +134,32 @@ export default class DiscordAdminRequest extends DiscordBasePlugin {
...
+    let amountAdmins = 0;
+    if (this.options.warnInGameAdmins) {
+      for (const player of this.server.players) {
+        if (!admins.includes(player.steamID)) continue;
+        amountAdmins++;
+        await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
+      }
+    }
+
+    if (amountAdmins === 0 && this.options.showInGameAdmins)
```
Thomas-Smyth: I don't think this works. As it stands current `amountAdmins` will always be zero if `warnInGameAdmins` is disabled. You're better removing the first if statement and reading it on the warn, like this:

```js
    let amountAdmins = 0;
    for (const player of this.server.players) {
      if (!admins.includes(player.steamID)) continue;
      amountAdmins++;
      if (this.options.warnInGameAdmins) await this.server.rcon.warn(player.steamID, `[${info.player.name}] - ${info.message}`);
    }
```

Thomas-Smyth (changes requested):

squad-server/plugins/discord-admin-request.js
LeventHAN: oh dang my bad 😞

PR #167, opened 2021-03-07
Add event broadcasts to the SocketIOAPI plugin

LeventHAN:
### Example: 
image

### DIFF
--- a/squad-server/plugins/socket-io-api.js
+++ b/squad-server/plugins/socket-io-api.js
@@ -3,6 +3,31 @@ import { Server } from 'socket.io';
 
 import BasePlugin from './base-plugin.js';
 
+const eventsToBroadcast = [
+  'CHAT_MESSAGE',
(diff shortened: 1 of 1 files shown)


Thomas-Smyth:
I think this PR is actually fine as it... I think the undefined issue probably relates to your clientside code.

For example, if you did:
```js
socket.on('UPDATED_PLAYER_INFORMATION', (info) => { console.log(info); });
```

Then this will return `undefined` because by design `UPDATED_PLAYER_INFORMATION` sends no information - No information is associated with this event because it's intended that the event listeners access the updated information directly via `server.players`. The code above expects there to be at least one argument, which isn't the case for this event, so it prints `undefined`. I suggest, if you're looping over a list of events, you also add the event name in the `console.log` message so you can see which events it's coming from and thus whether the `undefined is expected.

I'll leave this open for now until you confirm you're happy.

LeventHAN:
You are totally right...

PR #168, opened 2021-03-08
Fix seeding messages appearing after a map change whilst the server is live

LeventHAN:
Waiting for the previus PR to be accepted. (@Thomas-Smyth let me know if I need to close this one since I see the previus PR inside this one too, but perhaps it will dissapear after you accept the previus one dunno tbh)


LeventHAN:
Fixing https://github.com/Thomas-Smyth/SquadJS/issues/43

PR #184, opened 2021-03-16
Add number of helicopters and tanks per team to layer information

SeanWalsh95:
simplified to reduce commit spam

### DIFF
--- a/squad-server/layers/layer.js
+++ b/squad-server/layers/layer.js
@@ -15,33 +15,33 @@ export default class Layer {
       name: data.lighting,
       classname: data.lightingLevel
     };
-    this.teams = [
-      {
-        faction: data.team1.faction,
-        name: data.team1.teamSetupName,
-        tickets: data.team1.tickets,
-        commander: data.team1.commander,
-        vehicles: (data.team1.vehicles || []).map((vehicle) => ({
+    this.teams = [];
+    this.tanks = [];
+    this.helicopters = [];
+    for (const t of ['team1', 'team2']) {
+      this.teams.push({
+        faction: data[t].faction,
+        name: data[t].teamSetupName,
+        tickets: data[t].tickets,
+        commander: data[t].commander,
+        vehicles: (data[t].vehicles || []).map((vehicle) => ({
           name: vehicle.type,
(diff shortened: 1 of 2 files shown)


squad-server/layers/layer.js
```diff
@@ -15,33 +15,33 @@ export default class Layer {
...
-    this.teams = [
-      {
-        faction: data.team1.faction,
-        name: data.team1.teamSetupName,
-        tickets: data.team1.tickets,
-        commander: data.team1.commander,
-        vehicles: (data.team1.vehicles || []).map((vehicle) => ({
+    this.teams = [];
+    this.tanks = [];
+    this.helicopters = [];
```
Thomas-Smyth: Would it not make more sense for this to be a sub-property of the team? For example:
```js
      this.teams.push({
        faction: data[t].faction,
        name: data[t].teamSetupName,
        tickets: data[t].tickets,
        commander: data[t].commander,
        vehicles: (data[t].vehicles || []).map((vehicle) => ({
          name: vehicle.type,
          classname: vehicle.rawType,
          count: vehicle.count,
          spawnDelay: vehicle.delay,
          respawnDelay: vehicle.respawnTime
        })),
       numberOfHelicopters: // ...
      });
```

Thomas-Smyth (commented):
What's the use case for this? If it's not going to be used much might it just be better to do it in the plugin it's required?

SeanWalsh95:
> What's the use case for this? If it's not going to be used much might it just be better to do it in the plugin it's required?

Use case would be for providing a more rich set of data for any future plugin or filtering system looking at layer info, this was intended to provide similar information that the old layer/filter system provided in V1.

`layers.json` from V1
```
...
        "tanks": "N/A",
        "helicopters": "N/A",
...

        "tanks": "N/A",
        "helicopters": "x1 for GB, 20min delay",
...
```

Examples of uses could range from more info in the server status plugin to show next layer info as well as more info about the current layer, to bringing back a possible filtering system

squad-server/layers/layer.js
SeanWalsh95: sure I was working off the old `layers.json` model but this makes more sense

PR #190, opened 2021-04-07
Fix player connection event.

wiilover22:
It would appear the Squad log added some spaces, which was messing up with the regex for steamid-connected, which in return stopped player-connected from working as well. 

This appears to fix the issue, regex now matches the line in the log.


milutinke:
I've posted a fix on discord yesterday after some debugging.
I was planning to push a fix when my current pull request gets merged.
While you are at it, change this too (I already fixed the regex, just add it to the commit):

player-disconnected.js:
```
export default {
  regex: /^\[([0-9.:-]+)]\[([ 0-9]*)]LogEasyAntiCheatServer: \[[0-9:]+] \[[A-z]+] \[EAC Server] \[Info] \[UnregisterClient] Client: ([A-z0-9]+) PlayerGUID: ([0-9]{17})/,
  onMatch: (args, logParser) => {
    const data = {
      raw: args[0],
      time: args[1],
      chainID: args[2],
      steamID: args[4]
    };

    logParser.emit('PLAYER_DISCONNECTED', data);
  }
};
```

Thomas-Smyth:
I'll merge this and then let anyone open a PR for the disconnect message if it needs fixing too.

PR #192, opened 2021-04-19
Logging of roles in DB (Wounds+Deaths) + fix weapon bug @mattjons0n 

LeventHAN:
This PR includes;

- Fix for the weapons being shown as roles in DBLog_Deaths (Thanks to @mattjons0n (Jons0n#2156))
- Logging of roles in `DBLog_Wounds` & `DBLog_Deaths`

Servers that already have a running DB please run those commands in your MySQL server after updating your squadJS to this version;
```sql
-- Wounds
ALTER TABLE `DBLog_Wounds`
ADD COLUMN `role` VARCHAR(255) NULL DEFAULT NULL AFTER `weapon`;
```
```sql
-- Deaths
ALTER TABLE `DBLog_Deaths`
ADD COLUMN `role` VARCHAR(255) NULL DEFAULT NULL AFTER `weapon`;
```


squad-server/plugins/db-log.js
```diff
@@ -503,6 +509,7 @@ export default class DBLog extends BasePlugin {
       attackerSquadID: info.attacker ? info.attacker.squadID : null,
       damage: info.damage,
       weapon: info.weapon,
+      role: info.role,
```
Thomas-Smyth, line 512: How are you getting the role here if it's only in the die event log line?

Not sure I'm a fan of storing this information given it's very poor, i.e. only present on the die event. Probably better to get it another way, e.g. via Possess log lines, or wait until OWI add something.

Thomas-Smyth (changes requested):

squad-server/plugins/db-log.js
LeventHAN: You are right... It only works for Deaths and not Wounds 👍 
For now we may just log the role in Deaths and wait for OWI

I didn't realise that at first place thought it would just grab it from the Deaths, my bad for not testing it at first place. Sorry 😞

Thomas-Smyth: Have you looked at what the possess log line provides? I think the info might be stored in `server.players` too.

LeventHAN:
I believe I did broke this pr...

Will close this and re-open when I'm ready with this.

PR #193, opened 2021-05-17
Potential Fix to Warn Spam

werewolfboy13:
This fix should correct the issue if a player is in a squad and prevent them from getting warned. It appears to be working on our server.

### DIFF
--- a/squad-server/plugins/auto-kick-unassigned.js
+++ b/squad-server/plugins/auto-kick-unassigned.js
@@ -217,7 +217,7 @@ export default class AutoKickUnassigned extends BasePlugin {
         this.verbose(2, `Warning: ${tracker.player.name} (${timeLeft})`);
       }
       tracker.warnings++;
-   }, this.warningInterval);
+    }, this.warningInterval);
 
     // set timeout to kick player
     tracker.kickTimerID = setTimeout(async () => {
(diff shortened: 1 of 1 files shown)


LeventHAN (commented):
Would it not be enough to just put this to head of the function `if(!this.trackedPlayers.includes(tracker.player.steamID)) return;`
Regarding the speed your solution would be better but that would not fix the issue I believe...

Thomas-Smyth:
Odd this is being called... I'd have thought the timeout would have been removed when the player is untracked: https://github.com/Thomas-Smyth/SquadJS/blob/8b960ce7eb74bf10307dfd097b4b3717be27b53d/squad-server/plugins/auto-kick-unassigned.js#L245

werewolfboy13:
It warns one time then stops warning if they are in a squad.

LeventHAN:
> It warns one time then stops warning if they are in a squad.

So do I understand correctly?
1 - Players is without squad.
2 - Gets warned multiple times, until he joins a squad.
3 - Player joins a squad
4 - Gets still warned, **but** only one time...

werewolfboy13:
Correct how it was explained to me is that after it warns one more time it will stop then. It does warn then drops the flag/tracker.

LeventHAN:
> Correct how it was explained to me is that after it warns one more time it will stop then. It does warn then drops the flag/tracker.

Alright so after testing around. I found the solution.
**Issue**
When players join squad, the list of already tracking players is not getting updated. Which results in gettin another warning when player is still in squad. Until it updates (after 30 sec basiclly, but meanwhile the player is already in squad so gets only 1 time warning)

**To fix that**
You need to force update the tracking list when someone emits onPlayerSquadChange...

So basicly just do this:
```js
  async onPlayerSquadChange(player) {
    await this.updateTrackingList(true); // This forces to update the tracking list
    if (player.steamID in this.trackedPlayers && player.squadID !== null)
      this.untrackPlayer(player.steamID);
  }
```

werewolfboy13:
I have a WIP for this when I have time I will update it/

PR #205, opened 2021-08-10
Add execute

LeventHAN:
Will allow socketIO to execute all possible RCON commands from remote

### DIFF
--- a/squad-server/rcon.js
+++ b/squad-server/rcon.js
@@ -189,4 +189,9 @@ export default class SquadRcon extends Rcon {
   async switchTeam(steamID) {
     await this.execute(`AdminForceTeamChange "${steamID}"`);
   }
+  
+  // Will allow socketIO to execute direct rcon commands
+  async execute(...args) {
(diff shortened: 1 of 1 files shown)


Thomas-Smyth:
I don't get the point of this pull request, the child class will inherit the super class' `execute` method so why redefine it to what it's already defined as?

LeventHAN:
Well if you remember I did try to execute it (rcon.execute) however it was not working cz the super class... Therefor I must do this, as you also confirmed it would work like this.

I will double check this to be 200% sure 👍

11TStudio:
Well if you remember I did try to execute it (rcon.execute) however it was not working cz the super class... Therefor I must do this, as you also confirmed it would work like this.

I will double check this to be 200% sure 👍

LeventHAN:
Converting to draft until I'm able to test the stuff we spoke via PM

Thomas-Smyth:
Going to close this for now.

PR #206, opened 2021-08-10
Add DiscordKillFeed plugin

LeventHAN:
All changes have been tested, after waiting for 30 minutes on full server I never got a rateLimit (which did suprise me).

After looking the discord docs I have seen that the ratelimit for sending to DiscordAPI is 50req/s.

Fixing this: https://github.com/Thomas-Smyth/SquadJS/issues/197

### DIFF
--- a/squad-server/plugins/discord-killfeed.js
+++ b/squad-server/plugins/discord-killfeed.js
@@ -9,7 +9,7 @@ export default class DiscordKillFeed extends DiscordBasePlugin {
   }
 
   static get defaultEnabled() {
-    return true;
+    return false;
   }
 
   static get optionsSpecification() {
--- a/squad-server/plugins/discord-killfeed.js
+++ b/squad-server/plugins/discord-killfeed.js
@@ -1,0 +1,96 @@
+import DiscordBasePlugin from './discord-base-plugin.js';
+
+export default class DiscordKillFeed extends DiscordBasePlugin {
+  static get description() {
+    return (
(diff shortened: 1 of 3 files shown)


LeventHAN:
Hmm after 30 minutes, just now, there was a peak moment where a lot of wounds happend. Which basicly did try to send 6-7 wounds at once/same moment.
This let to hit the Discord's ratelimits.
Happened twice. 
Rate limit hit 4185
Rate limit hit 4298

But I believe the embeds are getting sended anyways... Not 100% sure tho.

LeventHAN:
As a side note;
Apparently 50mes/s is the global rate limit. On server limit it is **5mes/s** 😭

LeventHAN:
DiscordJS is apparently dealing with it and should not be a big problem.

squad-server/plugins/discord-killfeed.js
```diff
@@ -0,0 +1,96 @@
...
+export default class DiscordKillFeed extends DiscordBasePlugin {
+  static get description() {
+    return (
+      'The <code>DiscordKillFeed</code> plugin logs all wounds and related information to a Discord channel for ' +
+      'admins to review.'
+    );
+  }
+
+  static get defaultEnabled() {
+    return true;
```
Thomas-Smyth: I doubt many people will want to use this plugin, I would suggest setting the default to `false`.

Thomas-Smyth (changes requested):

squad-server/plugins/discord-killfeed.js
LeventHAN: Done 👍

PR #217, opened 2021-10-31
Fix missing teamkill event on instakill

MatthiasKunnen:
When a player got instakilled, due to large caliber or during grayed-out stage after revival, the teamkill event was not triggered.

Example from the logs:
```logs
[2021.10.31-00.38.35:319][ 83]LogSquad: Player:Victim ActualDamage=900.000000 from Attacker caused by BP_BTR82A_RUS_2A72_AP_C_2147479210
[2021.10.31-00.38.35:319][ 83]LogSquadTrace: [DedicatedServer]ASQSoldier::Die(): Player:Victim KillingDamage=900.000000 from BP_PlayerController_C_2147481608 caused by BP_BTR82A_RUS_2A72_AP_C_2147479210
```

### DIFF
--- a/squad-server/index.js
+++ b/squad-server/index.js
@@ -244,6 +244,7 @@ export default class SquadServer extends EventEmitter {
 
     this.logParser.on('PLAYER_DIED', async (data) => {
       data.victim = await this.getPlayerByName(data.victimName);
+      data.attacker = await this.getPlayerByName(data.attackerName);
 
       if (data.victim && data.attacker)
         data.teamkill =
@@ -254,6 +255,10 @@ export default class SquadServer extends EventEmitter {
       delete data.attackerName;
 
       this.emit('PLAYER_DIED', data);
+
+      if (data.teamkill) {
+        this.emit('TEAMKILL', data);
+      }
     });
 
     this.logParser.on('PLAYER_REVIVED', async (data) => {


squad-server/index.js
```diff
@@ -254,6 +255,10 @@ export default class SquadServer extends EventEmitter {
       delete data.attackerName;
 
       this.emit('PLAYER_DIED', data);
+
+      if (data.teamkill) {
+        this.emit('TEAMKILL', data);
```
Thomas-Smyth, line 260: Using the "TEAMKILL" event name is problematic because it also occurs when someone is wounded (iirc), thus this change could trigger somewhat duplicate events - not great considering people use this event within plugins that automatically ban players after x teamkill events.

Need to make changes to prevent that somehow.

Thomas-Smyth (changes requested):

squad-server/index.js
MatthiasKunnen: According to my testing, the following events exist:
- PLAYER_DAMAGED: the player is hurt but not down
- PLAYER_WOUNDED: the player is downed but revivable
- PLAYER_DIED: the player cannot be revived and must respawn

PLAYER_WOUNDED and PLAYER_DIED are mutually exclusive and preceded by the PLAYER_DAMAGED event.

Currently, only PLAYER_WOUNDED causes a TEAMKILL event and PLAYER_DIED does not.

### Examples

#### Multiple damage events leading to a TK, player is revivable
```log
[2021.10.31-01.13.44:364][426]LogSquad: Player:Watson ActualDamage=26.040001 from Charlz caused by BP_M4_Classic_M68_C_2147477285
[2021.10.31-01.13.45:125][464]LogSquad: Player:Watson ActualDamage=62.000004 from Charlz caused by BP_M4_Classic_M68_C_2147477285
[2021.10.31-01.13.45:125][464]LogSquadTrace: [DedicatedServer]ASQSoldier::Wound(): Player:Watson KillingDamage=62.000004 from BP_PlayerController_C_2147478566 caused by BP_M4_Classic_M68_C_2147477285
```

#### TK victim dies
Caused by instakills such as large caliber and greyed-out kills
```log
[2021.10.31-00.38.35:319][ 83]LogSquad: Player:Victim ActualDamage=900.000000 from Attacker caused by BP_BTR82A_RUS_2A72_AP_C_2147479210
[2021.10.31-00.38.35:319][ 83]LogSquadTrace: [DedicatedServer]ASQSoldier::Die(): Player:Victim KillingDamage=900.000000 from BP_PlayerController_C_2147481608 caused by BP_BTR82A_RUS_2A72_AP_C_2147479210
```

Thomas-Smyth: Do you mean PLAYER_WOUNDED and PLAYER_DIED are mutually exclusive? Looks like a typo?

MatthiasKunnen: > Do you mean PLAYER_WOUNDED and PLAYER_DIED are mutually exclusive? Looks like a typo?

oops, yes, exactly

Thomas-Smyth: They're not mutually exclusive, both can be trigger of a player is wounded and then gives us (or bleeds out afaik).

MatthiasKunnen: Just tested it and giving up/bleeding out triggers PLAYER_DIED indeed. 

I believe we can fix this by checking whether the damage event has the same chainID as the PLAYER_DIED event. what do you think? If it is not the same, it's not an instakill.

Thomas-Smyth: That's a possible solution, but not sure how much of a fan I am of the `TEAMKILL` event possibly being triggered at two different times when a player is wounded or killed. I think there need to be two separate events, e.g. `WOUNDED_TEAMKILL` and `DIED_TEAMKILL`, or just not use the events entirely and listen to the `PLAYER_WOUNDED` and `PLAYER_DIED` events and look at the teamkill flag in that.

Thomas-Smyth:
After removing the teamkill changes and testing this should be ready to merge.

PR #219, opened 2021-11-15
mapvote

bravo2zero:


ect0s:
https://regex101.com/r/e0Ui0K/1

Possible regex for new game. Shouldn't fail on vanilla maps, CAF etc. 

Works for all Vanilla modes, will capture maps correctly for mods, so long as there is a folder between the root and the layers. But we will capture that folder as the mapClassname

mapClassName is used in DBlog plugin. 

Mods are not bound to put the mapClassname (base layer) anywhere in the path when cooking their mods. They could have any folder structure, apart from the first entry and last entry before 'up at...'
```
LogWorld: Bringing World /(modfolder)/......./(layerfile.layerpersistentlevel) up at...
```

werewolfboy13:
Is there anything else that needs to be tested/worked with this?

bravo2zero:
> Is there anything else that needs to be tested/worked with this?

current functionality should work as it is, apart from merging/conflict resolving it should be good. I stopped working on this a while ago for the lack of time and also expecting v2.12 update to drop some changes.

Coolisbetter:
I would love to see this pull request merged.

bravo2zero:
closing this one, tested with v2.16 and moved to PR #241

PR #224, opened 2022-01-18
Suicide Handler

LeventHAN:
@Thomas-Smyth regarding DBLog I've just used onPlayerDied. Issue might be where attacker (and woundtime, etc) will be null.
The main purpose is there, I don't see any issue of the other info being null as the purpose is catching someone using suicide event and logging it/emitting it.

I don't think I miss anything, but double check please :)

### DIFF
--- a/squad-server/log-parser/player-suicide.js
+++ b/squad-server/log-parser/player-suicide.js
@@ -1,0 +1,17 @@
+export default {
+  regex: /^\[([0-9.:-]+)]\[([ 0-9]*)]LogSquad: Warning: Suicide (.+)/,
+  onMatch: (args, logParser) => {
+    const data = {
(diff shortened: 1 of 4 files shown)


squad-server/log-parser/player-suicide.js
```diff
@@ -0,0 +1,17 @@
+export default {
+  regex: /^\[([0-9.:-]+)]\[([ 0-9]*)]LogSquad: Warning: Suicide (.+)/,
```
werewolfboy13, line 2: Is this still valid with the current version of squad?

LeventHAN: > Is this still valid with the current version of squad?

I don't have any clue as I didn't touch squad for ages now :/

PR #225, opened 2022-02-09
Update for Squad v2.12

bumbummen99:
Add creator `name` & `steamId` properties to squad object,
Add `leader` and `role` properties to player object,
Add event `PLAYER_ROLE_CHANGED`
Add event `PLAYER_SQUADLEADER_GAINED`
Add event `PLAYER_SQUADLEADER_LOST`

Contains #226

### DIFF
--- a/squad-server/index.js
+++ b/squad-server/index.js
@@ -359,6 +359,26 @@ export default class SquadServer extends EventEmitter {
             oldSquadID: oldPlayerInfo[player.steamID].squadID,
             newSquadID: player.squadID
           });
+        if (
+          player.leader !== oldPlayerInfo[player.steamID].leader &&
+          oldPlayerInfo[player.steamID].leader === true
+        )
+          this.emit('PLAYER_SQUADLEADER_LOST', {
+            player: player
+          });
+        if (
+          player.leader !== oldPlayerInfo[player.steamID].leader &&
+          oldPlayerInfo[player.steamID].leader === false
+        )
+          this.emit('PLAYER_SQUADLEADER_GAINED', {
+            player: player
+          });
+        if (player.kit !== oldPlayerInfo[player.steamID].kit)
+          this.emit('PLAYER_ROLE_CHANGED', {
+            player: player,
+            oldRole: oldPlayerInfo[player.steamID].role,
+            newRole: player.role
+          });
       }
 
       this.emit('UPDATED_PLAYER_INFORMATION');
--- a/squad-server/index.js
+++ b/squad-server/index.js
@@ -290,11 +290,10 @@ export default class SquadServer extends EventEmitter {
     });
 
     this.logParser.on('SQUAD_CREATED', async (data) => {
-
-      data.player = await this.getPlayerBySteamID(data.playerSteamID, true)
-      delete data.playerName
-      delete data.playerSteamID
-      delete data.squadID
+      data.player = await this.getPlayerBySteamID(data.playerSteamID, true);
+      delete data.playerName;
+      delete data.playerSteamID;
+      delete data.squadID;
 
       this.emit('SQUAD_CREATED', data);
     });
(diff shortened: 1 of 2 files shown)


werewolfboy13 (approved):

bumbummen99:
@werewolfboy13 Are there any changes in regards to plugins that can be made to improve this PR? So far the data has only been provided, it is not being used.

ect0s:
I feel like the new Info would allow for an SL tracking plugin (which I started on, and I know others have as well)

DBLog for kit tracking. Although this will require a migrate script for the schema

Could also add kits to various discord plugins (IE, Discord Chat)


Depending on how much info people want sent to discord, the SL Gained/Lost could also be good in an audit log, I feel like really we should try to coalesce the change so it shows who gave up SL and who Gained SL, Requires more tinkering and prehaps a new event (SQUAD_SL_CHANGED?)

It might be useful to push a diff of [Old SL, New SL] so people can fireoff a nice single message.

Thomas-Smyth:
Just some initial feedback from a quick rear on my phone:
- Can we reword the event names please to read better, e.g. `PLAYET_PROMOTED_TO_SQUAD_LEADER`.
- It would perhaps be nice to have an event that's centered around the squad too, e.g. `SQUAD_LEADER_CHANGED`, but no rush on that... Might be better suited for the next major SquadJS version.

Thomas-Smyth:
> DBLog for kit tracking. Although this will require a migrate script for the schema

I'd be open to trying to migrate to Prisma... I played with it the other day and it seems pretty good in regards to migrations.

ect0s:
> Just some initial feedback from a quick rear on my phone:
> 
>     * Can we reword the event names please to read better, e.g. `PLAYET_PROMOTED_TO_SQUAD_LEADER`.
> 
>     * It would perhaps be nice to have an event that's centered around the squad too, e.g. `SQUAD_LEADER_CHANGED`, but no rush on that... Might be better suited for the next major SquadJS version.

PLAYER_SQUADLEADER_GAINED => PLAYER_PROMOTED_SQUAD_LEADER?
PLAYER_SQUADLEADER_LOST => PLAYER_DEMOTED_SQUAD_LEADER?


I used generic language because without more tracking we won't know if someone just created a squad, voluntarily passed SL, gave up a squad etc. 

Needs some thought to see how much context we can snag. 

As far as 

>     * It would perhaps be nice to have an event that's centered around the squad too, e.g. `SQUAD_LEADER_CHANGED`, but no rush on that... Might be better suited for the next major SquadJS version.


Do we track this in "Updated Squads" or "Update Players" or is this new logic we call somewhere else? 

My issue right now is we have a bit of a data race with where things get updated; Player Update Events vs Squad Update Events. I'm also just getting my coffee, so if the solution is obvious I apologize.  

Also need to test what happens at round end/round start as far as events firing off, I know v4 addresses shutting down some events during map transitions related to RCON, so perhaps better suited there.

Thomas-Smyth:
`PLAYER_BECAME_SQUAD_LEADER` sounds alright and addresses the issue you made about it wrongly implying the person was promoted when they actually just made the squad, but I'm not sure what the opposite of `BECAME` would be.

I was thinking it would be nice to have an alias of this event, triggered in the same place, that rather than containing the `player` object contains the `squad` object as it would make sense for dealing with things from a squad oriented perspective, e.g. if you were making a plugin to list all squads and their SLs in Discord. Either way, we can just leave it for now as it's only a nice to have.

ect0s:
Was giving this some thought today.

Perhaps it is more sane to simplify events:

PLAYER_ROLE_CHANGED [Oldplayer, Newplayer]

And then only fire SL changes from 

SQUAD_LEADER_CHANGED [OldSquad, NewSquad]

This eliminates the SL specific events, which helps maybe with naming etc; we don't directly address the issues of "became" vs "gained" vs "lost" but the information is still accessible, albeit a bit indirectly. 

the SL status events are just specific cases of role changes, so while its nice to directly access them, all the logic people would write based off these events can be fired off either of the above. 

I was thinking about how discord handles role updates and such.

ect0s:
https://github.com/ect0s/SquadJS/blob/1a53bcd8ccf3c2f278a032d47f7395865a2672a4/squad-server/index.js#L383
Not the most elegant code, but 

SQUAD_LEADER_CHANGE is implemented

nickbug:
Any update?

I want to write a leader kit require warning plugin, but it need leader and role properties in player object

before i open a new PR, I find this PR :(

werewolfboy13:
Cleaning up PRs is this still needing approved?

PR #226, opened 2022-02-11
Add Leader/Kit Change events pending RCON Changes Pending https://github.com/bumbummen99/SquadJS/tree/patch-3

ect0s:
Adds events based off new RCON information in

https://github.com/bumbummen99/SquadJS/tree/patch-3

PLAYER_SQUADLEADER_GAINED
PLAYER_SQUADLEADER_LOST
PLAYER_ROLE_CHANGED


bumbummen99:
You can PR/commit it into my branch, that way it will have the correct history in case someone does checkout the commit. Added you as collab

PR #229, opened 2022-02-12
Add Visual Studo .vs/ folder to .gitignore

werewolfboy13:
This will prevent the files the Visual Studio creates when editing the repo.


bumbummen99:
We could, for vscode at least, add an machine independent workspace and launch json @werewolfboy13 what do you think?

PR #240, opened 2022-05-27
Add default value for victim's name in the DiscordKillFeed plugin

ect0s:


ect0s:
```
[2022.04.09-07.06.55:853][985]LogSquadTrace: [DedicatedServer]Wound(): Player:nullptr KillingDamage=176.295227 from BP_PlayerController_C_2146878680 caused by BP_BTR82A_RUS_2A72_AP_Desert_C_2146855463
[2022.04.09-07.19.59:417][898]LogSquadTrace: [DedicatedServer]Wound(): Player:nullptr KillingDamage=258.000000 from BP_PlayerController_C_2147165493 caused by BP_MG3_DoorGun_C_2146825055
[2022.04.09-07.20.49:605][150]LogSquadTrace: [DedicatedServer]Wound(): Player:nullptr KillingDamage=62.000004 from BP_PlayerController_C_2146904499 caused by BP_Minimi_MK1_SCROMEJ4_C_2146823536
[2022.04.09-07.22.52:064][639]LogSquadTrace: [DedicatedServer]Wound(): Player:nullptr KillingDamage=185.000000 from BP_PlayerController_C_2147129998 caused by BP_DRAGAR_GIAT_M811_AP_Desert_C_2146824927
```

README.md
```diff
@@ -1,7 +1,6 @@
 <div align="center">
 
-  <img src="assets/squadjs-logo-white.png#gh-dark-mode-only" alt="Logo" width="500"/>
-<img src="assets/squadjs-logo.png#gh-light-mode-only" alt="Logo" width="500"/>
```
Thomas-Smyth: It looks like this change was a mistake?

Thomas-Smyth:
As a side note, do we know why `victim` not being defined is being reported a lot more recently than it has in the past? Is it just because this plugin is relatively new and doesn't have handling for it, unlike the others?

README.md
ect0s: I didn't change that intentionally. 

I also thought I pulled from squadjs latests.. I wonder if something in the linter/yarn scripts is acting up on my end, I've noticed whitespace changes after each commit for a while.

ect0s:
> As a side note, do we know why `victim` not being defined is being reported a lot more recently than it has in the past? Is it just because this plugin is relatively new and doesn't have handling for it, unlike the others?

I believe within the DBlog plugin unassigned values or missing properties are coerced to null by Sequalize; though I may be incorrect. 

Discordjs, however does not; especially when the code in question is more or less just constructing an object in the raw and passing it along (so, I guess this could be phrased as nodejs itself?) 

It is interesting.


As an aside, I've noticed many people are using newer versions of nodejs (>14) against the install guide, and this has made errors involving promise rejection halt the process which is much more noticable

README.md
Thomas-Smyth: Possibly, can you try reverting it, please?

ect0s: I think this is husky rebuilding the readme file, checking to confirm.

I'll need to troubleshoot a bit, but in testing right now on a fresh branch, the moment I ran a commit the changes appeared.

Thomas-Smyth:
> As an aside, I've noticed many people are using newer versions of nodejs (>14) against the install guide, and this has made errors involving promise rejection halt the process which is much more noticable

To be fair, the reason for using that version of Node is probably no longer relevant. But yes, fixing the crashing is something that needs to be looked into - probably in the new major version as it might require larger structural changes to be robust.

ect0s:
If we do a version upgrade, currently we are missing 

https://github.com/discordjs/discord.js/commit/9d2d60691eb4bde729f40fb633ae257cf5bc6545

From discordjs, which may explain another issue with the above plugin. Someone had mentioned on SQ RCON discord that it seamed they were getting rate limited on a channel or whatever, requiring a restart, but I'm hazy on the details.

README.md
Thomas-Smyth: Try merging in master and trying again please - it looks like I forgot to add the light logo to the README template.

PR #243, opened 2022-06-04
Move sourcing of SQUAD_CREATED event to RCON

ect0s:


werewolfboy13:
@ect0s can you resolve the conflicts?

squad-server/index.js
```diff
@@ -152,6 +152,15 @@ export default class SquadServer extends EventEmitter {
 
       this.emit('PLAYER_BANNED', data);
     });
+
+    this.rcon.on('SQUAD_CREATED', async (data) => {
+      data.player = await this.getPlayerBySteamID(data.playerSteamID, true);
+      delete data.playerName;
+      delete data.playerSteamID;
+      delete data.squadID;
```
Thomas-Smyth: Do we want to wipe this? It seems like good information to have.

Thomas-Smyth (dismissed):
LGTM, just need to fix the merge conflicts.

squad-server/index.js
ect0s: We could leave it. 

I wanted to move SQCREATED to RCON because it *should* be more reliable, especially for those on FTP logging. 

Its been a while since I looked at/tested this PR, but I believe theres no harm in leaving the data, especially when these events fire *often* on map change and theres a slight race with this.updatePlayers loop and the getPlayerBySteamID call.

Thomas-Smyth: I was mainly referring to `squadID` only, the rest makes sense to be inside the player object, just the `squadID` is more generic so is probably more relevant on the top level of the event.

ect0s: This should now be fixed. 

To fix the conflict I had to roll this branch forward.. I'm not sure I've done that correctly, reviewing

werewolfboy13: Conflict appears to be resolved? The file was not removed properly. 

@Thomas-Smyth can you make sure nothing broke?
image

PR #245, opened 2022-06-28
Replace SCBL integrations with CBL integrations

IgnisAlienus:
Updating the Wording and Links to reference to the new Community Ban List. Renamed scbl-info.js to cbl-info.js and updated the refences to SCBL in discord-killfeed.js and discord-teamkill.js

Looks like some of the automations also cleaned up some code in a few other files. You don't have to accept those.


Thomas-Smyth:
It doesn't look like any changes to the SCBL plugin were committed?

IgnisAlienus:
> It doesn't look like any changes to the SCBL plugin were committed?

Oooof, I'll double check when I get back to my PC. The scbl-info should be replaced with cbl-info.

IgnisAlienus:
Okay, got it now. 👌

PR #250, opened 2022-07-12
Auto TK Warn Victim

IgnisAlienus:
Added a plugin that will warn the TK Victim that they've been TK'd with the default message of "INFO: You were Teamkilled." This plugin is set to `false` by default in the config.json file.

I know it could be controversial to warn the TK Victim about being TK'd, but we at PSG are using this and haven't seen any difference in retaliatory TKs. It seems most blueberries get more satisfaction out of knowing that they were TK'd over wondering what possible enemy they missed in the treeline that just dunked on them when it was a TK all along.

Because of how some Servers may decide to not warn the TK Victim, that is the reason this is defaulted to be disabled.

It also does not tell the TK Victim who TK'd them, just that they were TK'd.

Updated the original TK Warn to Auto TK Warn Attacker to remove any possible confusion of this addition.


package.json
```diff
@@ -20,6 +20,7 @@
   },
   "type": "module",
   "dependencies": {
+    "config": "^3.3.7",
```
werewolfboy13: Why is this needed?

watchful-eyes:
It's an amazing idea but isn't better adding an option for warn filter such as
 ```
warnAttacker: true, 
 warnVictim: true
``` 
or putting into array
instead of making a whole another plugin

package.json
IgnisAlienus: It must of put it in when I ran the config build.

I'll edit it out.

IgnisAlienus: All set.

IgnisAlienus:
> It's an amazing idea but isn't better adding an option for warn filter such as
> 
> ```
> warnAttacker: true, 
> warnVictim: true
> ```
> 
> or putting into array instead of making a whole another plugin

I'm not super great at coding, this method of a new plugin was easiest for me to accomplish while getting the desired result. I'm all for it if someone else would like to make it into one plugin.

PR #251, opened 2022-07-12
Add option to warn victims of teamkills via AutoTKWarn plugin

SeanWalsh95:
alternate solution to PR #250

### DIFF
--- a/squad-server/plugins/auto-tk-warn.js
+++ b/squad-server/plugins/auto-tk-warn.js
@@ -16,11 +16,6 @@ export default class AutoTKWarn extends BasePlugin {
         description: 'The message to warn attacking players with.',
         default: 'Please apologise for ALL TKs in ALL chat!'
       },
-      warnVictim: {
-        required: false,
-        description: 'Should we nottify the the victim, telling them that they were teamkilled via a warning message',
-        default: false
-      },
       victimMessage: {
         required: false,
         description: 'The message that will be sent to the victim',
@@ -44,11 +39,11 @@ export default class AutoTKWarn extends BasePlugin {
   }
 
   async onTeamkill(info) {
-    if (info.attacker) {
-      this.server.rcon.warn(info.attacker.steamID, this.options.attackerMessage);
+    if (info.attacker && this.options.attackerMessage) {
+      await this.server.rcon.warn(info.attacker.steamID, this.options.attackerMessage);
     }
-    if (this.options.warnVictim && info.victim) {
-      this.server.rcon.warn(info.victim.steamID, this.options.victimMessage);
+    if (info.victim && this.options.victimMessage) {
+      await this.server.rcon.warn(info.victim.steamID, this.options.victimMessage);
     }
   }
 }
--- a/squad-server/plugins/auto-tk-warn.js
+++ b/squad-server/plugins/auto-tk-warn.js
@@ -11,17 +11,17 @@ export default class AutoTKWarn extends BasePlugin {
 
   static get optionsSpecification() {
     return {
-      attacker_message: {
+      attackerMessage: {
         required: false,
         description: 'The message to warn attacking players with.',
         default: 'Please apologise for ALL TKs in ALL chat!'
       },
-      warn_victim: {
+      warnVictim: {
         required: false,
         description: 'Should we nottify the the victim, telling them that they were teamkilled via a warning message',
         default: false
       },
-      victim_message: {
+      victimMessage: {
         required: false,
         description: 'The message that will be sent to the victim',
         default: 'You were killed by your own team.'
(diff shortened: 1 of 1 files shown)


squad-server/plugins/auto-tk-warn.js
```diff
@@ -11,10 +11,20 @@ export default class AutoTKWarn extends BasePlugin {
 
   static get optionsSpecification() {
     return {
-      message: {
+      attacker_message: {
```
Thomas-Smyth: This should be camel case, i.e. `attackerMessage`.

squad-server/plugins/auto-tk-warn.js
```diff
@@ -11,10 +11,20 @@ export default class AutoTKWarn extends BasePlugin {
...
         required: false,
-        description: 'The message to warn players with.',
+        description: 'The message to warn attacking players with.',
         default: 'Please apologise for ALL TKs in ALL chat!'
+      },
+      warn_victim: {
```
Thomas-Smyth: As above.

squad-server/plugins/auto-tk-warn.js
```diff
@@ -11,10 +11,20 @@ export default class AutoTKWarn extends BasePlugin {
...
+        required: false,
+        description: 'Should we nottify the the victim, telling them that they were teamkilled via a warning message',
+        default: false
+      },
+      victim_message: {
```
Thomas-Smyth: As above.

squad-server/plugins/auto-tk-warn.js
```diff
@@ -34,8 +44,11 @@ export default class AutoTKWarn extends BasePlugin {
   }
 
   async onTeamkill(info) {
-    if (!info.attacker) return;
-
-    await this.server.rcon.warn(info.attacker.steamID, this.options.message);
+    if (info.attacker) {
+      this.server.rcon.warn(info.attacker.steamID, this.options.attacker_message);
+    }
+    if (this.options.warn_victim && info.victim) {
```
Thomas-Smyth: We can probably do away with the `warn_victim` and just do it based on whether the message is defined or not - which means they do not need to be required.

```js
if (info.attacker && this.options.attackerMessage) {
  await this.server.rcon.warn(info.attackers.steamId, this.options.attackerMessage);
}

if (info.victim && this.options.victimMessage) {
  await this.server.rcon.warn(info.attackers.steamId, this.options.victimMessage);
}
```

Thomas-Smyth (changes requested):

squad-server/plugins/auto-tk-warn.js
SeanWalsh95: oops I've been writing python too long

squad-server/plugins/auto-tk-warn.js
SeanWalsh95: I was wondering is the await necessary? wouldn't one warning request just block the other until it gets a response?

Thomas-Smyth: Yeah, that's true, we can probably do without them.

squad-server/plugins/auto-tk-warn.js
```diff
@@ -11,10 +11,15 @@ export default class AutoTKWarn extends BasePlugin {
...
-      message: {
+      attackerMessage: {
         required: false,
-        description: 'The message to warn players with.',
+        description: 'The message to warn attacking players with.',
         default: 'Please apologise for ALL TKs in ALL chat!'
+      },
+      victimMessage: {
+        required: false,
+        description: 'The message that will be sent to the victim',
```
Thomas-Smyth: Can we add a full stop please?

Thomas-Smyth (changes requested):

squad-server/plugins/auto-tk-warn.js
```diff
@@ -11,10 +11,15 @@ export default class AutoTKWarn extends BasePlugin {
...
+      attackerMessage: {
         required: false,
-        description: 'The message to warn players with.',
+        description: 'The message to warn attacking players with.',
         default: 'Please apologise for ALL TKs in ALL chat!'
+      },
+      victimMessage: {
+        required: false,
+        description: 'The message that will be sent to the victim.',
+        default: 'You were killed by your own team.'
```
Thomas-Smyth: One last thing, I think we should default this to `null` so it's not enabled by default.

Thomas-Smyth (changes requested):

squad-server/plugins/auto-tk-warn.js
SeanWalsh95: yeah that was the original reason for the separate flag I'll set it to null and comment out the suggested message

IgnisAlienus:
I'm closing out my PR, #250 , for this since this one is here.

PR #256, opened 2022-09-29
Improve round end information, and add DiscordRoundEnded plugin

ect0s:
```

Adds round-tickets regex to logparser; this will fire when the server prints new log lines related to tickets and factions.

Adds round-ended regex to log parser and event; this event forwards round tickets where we have both a winner and a loser to the main server object.

discord-roundended: This plugin is similar but functionally different from round winner, due to cases where the server experiences a draw.
```

This is an initial commit, and will need additional work.

Currently known issues: 

We only see round-tickets when the map is not a draw, this is fairly often, however our current round-winner event *always* fires (though incorrectly labels all matches with a winner, even if they are a draw). 

The larger issue is the changes required in DBlog and elsewhere to smooth things over, these will result in schema changes that require DBlog to alter tables. 

Further, in the case of FTP logging, we currently assume the log and rcon state are closely coupled, this is not always the case; 

This means plugins like discord-roundwinner, that take the previous map are not guaranteed to be correct, and further issues around how we construct that data abound (layers.json missing data, see https://github.com/Team-Silver-Sphere/SquadJS/pull/222 )

The well defined path here should be stable, but edge cases around FTPlogging and such make the "draw" path much harder to handle. 

```
[2022.09.28-18.23.37:222][892]LogSquadTrace: [DedicatedServer]DetermineMatchWinner(): Local Militia Group won on Kamdesh Highlands
[2022.09.28-18.23.37:222][892]LogSquadTrace: [DedicatedServer]DetermineMatchWinner(): The game was a draw on Kamdesh Highlands
[2022.09.28-18.23.37:223][892]LogGameState: Match State Changed from InProgress to WaitingPostMatch

```

The above is a map with a draw, 

Currently NEW_GAME will report 

```

Local Militia Group won on Kamdesh Highlands
```

But this is incorrect, and the new log lines will not appear, leaving an ambiguous state.

### DIFF
--- a/squad-server/log-parser/round-ended.js
+++ b/squad-server/log-parser/round-ended.js
@@ -1,0 +1,20 @@
+/**
+ * Matches when Map state Changes to PostMatch (ScoreBoard)
+ *
+ * Emits winner and loser from eventstore
+ *
+ * winner and loser may be null if the match ends with a draw
+ */
+export default {
+  regex:
+    /^\[([0-9.:-]+)]\[([ 0-9]*)]LogGameState: Match State Changed from InProgress to WaitingPostMatch/,
+  onMatch: (args, logParser) => {
+    const data = {
+      winner: logParser.eventStore.ROUND_WINNER ? logParser.eventStore.ROUND_WINNER : null,
+      loser: logParser.eventStore.ROUND_LOSER ? logParser.eventStore.ROUND_LOSER : null
+    };
+    logParser.emit('ROUND_ENDED', data);
+    delete logParser.eventStore.ROUND_WINNER;
+    delete logParser.eventStore.ROUND_LOSER;
+  }
+};
--- a/squad-server/log-parser/round-ended.js
+++ b/squad-server/log-parser/round-ended.js
@@ -11,7 +11,8 @@ export default {
   onMatch: (args, logParser) => {
     const data = {
       winner: logParser.eventStore.ROUND_WINNER ? logParser.eventStore.ROUND_WINNER : null,
-      loser: logParser.eventStore.ROUND_LOSER ? logParser.eventStore.ROUND_LOSER : null
+      loser: logParser.eventStore.ROUND_LOSER ? logParser.eventStore.ROUND_LOSER : null,
+      time: args[1]
     };
     logParser.emit('ROUND_ENDED', data);
     delete logParser.eventStore.ROUND_WINNER;
--- a/README.md
+++ b/README.md
@@ -640,6 +640,29 @@ Grafana:
         </details>
 
 <details>
+          <summary>DiscordRoundEnded</summary>
+          <h2>DiscordRoundEnded</h2>
+          <p>The <code>DiscordRoundEnded</code> plugin will send the round winner to a Discord channel.</p>
+          <h3>Options</h3>
+          <ul><li><h4>discordClient (Required)</h4>
+           <h6>Description</h6>
+           <p>Discord connector name.</p>
+           <h6>Default</h6>
+           <pre><code>discord</code></pre></li>
+<li><h4>channelID (Required)</h4>
+           <h6>Description</h6>
+           <p>The ID of the channel to log round end events to.</p>
+           <h6>Default</h6>
+           <pre><code></code></pre></li><h6>Example</h6>
+           <pre><code>667741905228136459</code></pre>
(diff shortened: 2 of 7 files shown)


werewolfboy13:
Got this error on mine when I added it. 
image

ect0s:
The events are a composite set of two. 

Is it possible you restarted during the two events?

Or could you get me the relevant game log, and layer?

ect0s:
It appears I missed the early return for draws



> Got this error on mine when I added it. image

This should be fixed

ect0s:
```
[2022.09.29-10.47.59:424][474]LogSquadGameEvents: Display: Team 2, RUS 49th Combined Arms Army ( Russian Ground Forces ) has won the match with 294 Tickets on layer Narva RAAS v5 (level Narva)!
[2022.09.29-10.47.59:424][474]LogSquadGameEvents: Display: Team 1, USMC 3rd Light Armored Recon Battalion ( United States Marine Corps ) has lost the match with 296 Tickets on layer Narva RAAS v5 (level Narva)!
```

It appears the ticket totals we get may not always be super accurate. Squad has long had an issue of tickets getting subtracted during/after the scoreboard appears, I believe this is the issue under the hood. 

Will need additional testing to see how this occurs. The above was generated on a map that timed out after the test server emptied for the night

watchful-eyes:
> ```
> [2022.09.29-10.47.59:424][474]LogSquadGameEvents: Display: Team 2, RUS 49th Combined Arms Army ( Russian Ground Forces ) has won the match with 294 Tickets on layer Narva RAAS v5 (level Narva)!
> [2022.09.29-10.47.59:424][474]LogSquadGameEvents: Display: Team 1, USMC 3rd Light Armored Recon Battalion ( United States Marine Corps ) has lost the match with 296 Tickets on layer Narva RAAS v5 (level Narva)!
> ```
> 
> It appears the ticket totals we get may not always be super accurate. Squad has long had an issue of tickets getting subtracted during/after the scoreboard appears, I believe this is the issue under the hood.
> 
> Will need additional testing to see how this occurs. The above was generated on a map that timed out after the test server emptied for the night



Is that really the case?
Yes it inacurrate but it gets layer info directly from log and ticket counts doesn't have effect in code.

ect0s:
> Is that really the case?

Yes, the above log lines are directly from testing. 

In the past, before OWI added the 'victory flags' animations to the scoreboard, you would often see tickets getting adjusted slightly during the scoreboard screen. It was entirely possibly that before moving to the next match both teams would have negative tickets. 

> Yes it inacurrate but it gets layer info directly from log and ticket counts doesn't have effect in code.

The accuracy is an issue if anyone starts tracking ticket differentials. The data may be harder to understand if the winning team has less tickets logged than the loser.

The tickets don't have an effect in code until someone decides to write plugins off the event, or include them in Dblog  etc

werewolfboy13 (approved):
Tested on live server.

Sentennial (approved):
**Tested on private server.**
Three tests shown in order in SS below.

1. Match ended in draw
2. Match ended with 1 ticket difference
3. Match ended with 2 ticket difference

results

Coolisbetter (approved):
approve

edit: oops wrong account, lol

Sentennial (approved):
triage perms approve

squad-server/log-parser/round-ended.js
```diff
@@ -0,0 +1,21 @@
...
+ * Emits winner and loser from eventstore
+ *
+ * winner and loser may be null if the match ends with a draw
+ */
+export default {
+  regex:
+    /^\[([0-9.:-]+)]\[([ 0-9]*)]LogGameState: Match State Changed from InProgress to WaitingPostMatch/,
+  onMatch: (args, logParser) => {
+    const data = {
+      winner: logParser.eventStore.ROUND_WINNER ? logParser.eventStore.ROUND_WINNER : null,
```
Thomas-Smyth, line 13: You could do `logParser.eventStore.ROUND_WINNER || null` here I think.

Thomas-Smyth (approved):
LGTM, thanks for the PR.

squad-server/log-parser/round-ended.js
Sentennial: I think they both would have the same effect. I prefer ect0s' code imo. More generally readable I find.

PR #265, opened 2023-01-12
Update tedious package

DrKittens:
Latest version resolves a critical bug around xml dom.

Ref: https://github.com/advisories/GHSA-crh6-fp67-6883

Currently still testing impacts of this major version change but it hasnt broken my MariaDB connection for the app or SSJS so far. 

*Testing on RHEL9 **NOT WINDOWS** so could be OS specific breakage i'd miss.

This is probably low-priority for the project as i dont think we do anything with untrusted XML.

### DIFF
--- a/squad-server/package.json
+++ b/squad-server/package.json
@@ -17,7 +17,7 @@
     "sequelize": "^6.3.5",
     "socket.io": "^3.1.2",
     "sqlite3": "^5.0.0",
-    "tedious": "^9.2.1",
+    "tedious": "^15.1.2",
     "tinygradient": "^1.1.2"
   },
   "exports": {


werewolfboy13 (changes requested):
Needs more testing before PR can be pushed as this is a major version bump.

ect0s:
I think our dependence on Tedious is due to Sequelize, and its an optional dependency there if I'm reading yarn info correctly. 

@werewolfboy13 I know you may hesitate, but this should a low impact.

werewolfboy13:
> I think our dependence on Tedious is due to Sequelize, and its an optional dependency there if I'm reading yarn info correctly.
> 
> @werewolfboy13 I know you may hesitate, but this should a low impact.

I just want nothing to break, if it's optional should be fine. I'll push this on my internal tools and see what happens. This will occur in a couple days or so.

werewolfboy13 (approved):

ect0s (approved):

Issue #249, opened 2022-07-11
Possible Memory Leak 


mattjons0n:
### Description of Issue
Bot fails at doing any task over Rcon due to potential Memory leak

Error:  (node:3891673) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [Socket]. Use emitter.setMaxListeners() to increase limitError:  (Use `node --trace-warnings ...` to show where the warning was created)[LogParser][1] Lines parsed per minute: 807 lines per minute | Matching lines per minute: 186 matching lines per minute | Average matching latency: 1255.1290322580646ms

Output after above error:
[RCON][1] Unknown packet type 16777231 in: { size: 256, id: -1946157056, type: 16777231, body: '\x00\x00\x00\x00\x00\x00\x00----- Active Players -----\nID: 71 | SteamID: 76561199086558645 | Name: (   ÍœÍžÊ– ) | Team ID: 1 | Squad ID: 4 | Is Leader: False | Role: USA_Rifleman_03\nID: 75 | SteamID: 76561199259877501 | Name: 2945183**9 | Team ID: 1 | Squad ID: 4 | I' }

After above Unknown packet type output, no more Rcon log type 1 with player information is shown and below is looped.

[RCON][2] Writing packet with type "2" and body "ListPlayers".
[SquadServer][1] Updated A2S information.
[SquadServer][1] Updating player list...
[RCON][2] Writing packet with type "2" and body "ListPlayers".
[SquadServer][1] Updating player list...
[RCON][2] Writing packet with type "2" and body "ListPlayers".
[SquadServer][1] Updating player list...
[RCON][2] Writing packet with type "2" and body "ListPlayers".
[SquadServer][1] Updating player list...
[RCON][2] Writing packet with type "2" and body "ListPlayers".
[SquadServer][1] Updating player list...
[RCON][2] Writing packet with type "2" and body "ListPlayers".
[SquadServer][1] Updating player list...
[RCON][2] Writing packet with type "2" and body "ListPlayers".
[SquadServer][1] Updating player list...

### Squad Information
 * SquadJS Version: 3.4.0
 * Squad Version: 3.1.0

If potentially relevant, please provide regarding the state of the Squad server at the time of error, e.g. the current layer.

### System Information
 * Operating System: Ubuntu 20.04
 * NodeJS Version: 16.15.1
 * Yarn Version: 1.22.10

lbzepoqo:
[...]

Issue #262, opened 2023-01-07
RCON closes with error, no error stack


ect0s:
### Description of Issue
Every now and then RCON will close with an error, but without an error stack log line, 

```
[RCON][1] Disconnecting from: 15.235.160.4:21114
[RCON][1] Socket closed with an error.
[RCON][1] Disconnected from: 15.235.160.4:21114
[SquadServer][1] Setting up new RCON instance...
[RCON][1] Connecting to: 15.235.160.4:21114
[RCON][1] Connected to: 15.235.160.4:21114
[RCON][1] Authentication succeeded.
```
From here:
https://github.com/Team-Silver-Sphere/SquadJS/blob/dc2bc1d663fb5b4aecab39d9072ef51dad089fae/core/rcon.js#L180

I would expect to see Logger.verbose('RCON', 1, `Socket had error:`, err);

### Squad Information
 * SquadJS Version: 3.6.1
 * Squad Version: 4.1

If potentially relevant, please provide regarding the state of the Squad server at the time of error, e.g. the current layer.

### System Information
 * Operating System: Linux
 * NodeJS Version: Node14
 * Yarn Version: N/A

I believe this may be due to us missing an 'end' event from the TCP socket, where the squad server has closed its side for whatever reason; though I'm still confused as to why we never see the error surface.

I had noticed this in previous versions of squadjs, but generally when our server had died off in the night.

https://nodejs.org/api/net.html#event-end

https://github.com/Team-Silver-Sphere/SquadJS/blob/dc2bc1d663fb5b4aecab39d9072ef51dad089fae/core/rcon.js#L35

ect0s:
https://github.com/Team-Silver-Sphere/SquadJS/blob/dc2bc1d663fb5b4aecab39d9072ef51dad089fae/core/rcon.js#L172

Is this ternary backwards? 

Close() with out data will result in the false case; leading to "with an Error"

reck1610:
Yeah, that's backwards.

I checked the node documentation to rule out, that the parameter was poorly named.
https://nodejs.org/api/net.html#event-close_1
>Event: 'close'
> Added in: v0.1.90
> - `hadError` [\<boolean\>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type) true if the socket had a transmission error.
>
>Emitted once the socket is fully closed. The argument hadError is a boolean which says if the socket was closed due to a transmission error.

ect0s:
[...]

PR #267, opened 2023-01-16
Improve RCON packet Parsing and Cleanup on connection close to prevent deadlocks

### LINKED ISSUE #200: SquadRcon.onData Crashing Recently
### Description of Issue
Crashed twice in the past two days on the following callstack; One immediately after a server crash, so I assumed it was related to a missing packet from the crash. 

However, this morning we crashed shortly after hitting live during Fallujah Skimish v2

### Errors or Screenshots of Issue
file:///opt/SquadJS/core/rcon.js:80
              this.responseCallbackQueue.shift()(
                                                ^

TypeError: this.responseCallbackQueue.shift(...) is not a function
    at SquadRcon.onData (file:///opt/SquadJS/core/rcon.js:80:49)
    at Socket.emit (events.js:315:20)
    at addChunk (internal/streams/readable.js:309:12)
    at readableAddChunk (internal/streams/readable.js:284:9)
    at Socket.Readable.push (internal/streams/readable.js:223:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:188:23)
### Squad Information
 * SquadJS Version: v3
 * Squad Version: 2.9

If potentially relevant, please provide regarding the state of the Squad server at the time of error, e.g. the current layer.

### System Information
 * Operating System: Linux
 * NodeJS Version: Node 14
 * Yarn Version:

ect0s:
Restructure Packet Decode
Add counter for packets/callbacks
onClose() cleanups
unknown packet should call onClose(). 

This isn't production ready, but I felt the need to pull request it and solicit feedback. 

This is primarily to fix the issues as noted in https://github.com/Team-Silver-Sphere/SquadJS/issues/200 and https://github.com/Team-Silver-Sphere/SquadJS/issues/249

This should run as is, I have tested to above 200k RCON command requests, and across server restarts. Before these changes I could cause an error within the first few thousand requests. 

The error handling, in particular, could be improved, but I was trying to keep the code broadly similar to the current release. 

Longer term, a rewrite might incorporate the new 'count/id' field/callback tracking/retries into a single object, thus bringing some relevant code together

### DIFF
--- a/core/rcon.js
+++ b/core/rcon.js
@@ -54,6 +54,7 @@ export default class Rcon extends EventEmitter {
     // Used For tracking Callbacks
     this.callbackIds = [];
     this.count = 1;
+    this.loggedin = false;
   }
 
   onPacket(decodedPacket) {
@@ -181,7 +182,7 @@ export default class Rcon extends EventEmitter {
 
   onClose(hadError) {
     this.connected = false;
-
+    this.loggedin = false;
     Logger.verbose(
       'RCON',
       1,
@@ -306,6 +307,11 @@ export default class Rcon extends EventEmitter {
         return;
       }
 
+      if (!this.loggedin && type !== SERVERDATA_AUTH) {
+        reject(new Error('RCON not Logged in'));
+        return;
+      }
+
       Logger.verbose('RCON', 2, `Writing packet with type "${type}" and body "${body}".`);
 
       const encodedPacket = this.encodePacket(
@@ -328,8 +334,9 @@ export default class Rcon extends EventEmitter {
       };
 
       // the auth packet also sends a normal response, so we add an extra empty action to ignore it
-      this.callbackIds.push({ id: this.count, cmd: body });
+
       if (type === SERVERDATA_AUTH) {
+        this.callbackIds.push({ id: this.count, cmd: body });
         this.responseCallbackQueue.push(() => {});
         this.responseCallbackQueue.push((decodedPacket) => {
           this.client.removeListener('error', onError);
@@ -338,10 +345,12 @@ export default class Rcon extends EventEmitter {
             reject(new Error('Authentication failed.'));
           } else {
             Logger.verbose('RCON', 1, 'Authentication succeeded.');
+            this.loggedin = true;
             resolve();
           }
         });
       } else {
+        this.callbackIds.push({ id: this.count, cmd: body });
         this.responseCallbackQueue.push((response) => {
           this.client.removeListener('error', onError);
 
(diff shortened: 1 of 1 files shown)


lbzepoqo:
I've been using these changes without restarting my SquadJS and haven't encountered a single unknown packet since then.

Before, we were experiencing an issue where Admin Requests plugin, JetDave's map vote and Whitelister, along with other commands doesn't get read by SquadJS, but the process is still running and broadcasts were running fine. As I was investigating recently, we usually get 5 or more unknown packets within a single day and this is just when we notice that things doesn't work any more than we restart. As I discovered `process.abort()`, I included it in the code to crash our SquadJS instead of me manually restarting SquadJS and actively monitoring all day. In a span of 7-8 hours of active time, we get around 5+ of these unknown packets per day, and we had one even if the server wasn't full yet at 6/100.

My first encounter with this issue can be found in a Discord message [here](https://discord.com/channels/266210223406972928/714529190493290527/1017425294728110100). Since then, we've been running SquadJS to restart every ONE hour as suggested by Forrest Gump who also had an instance in Southnode where the SquadJS is located on a different server and is using FTP. They haven't experienced this when they went dedicated. I even requested for JetDave's AutomaticSubsystemRestarter as a workaround.

Also to mention, when an unknown packet appears in our log, the DiscordSubsystemRestarter is enough to fix the issue but still we had to do it manually. The problem with this remedy is that it doesn't restore the functionality of plugins using the websocket (for example: Whitelister's !profile command). There's also an issue where using the DiscordSubsystemRestarter or JetDave's AutomaticSubsystemRestarter sometimes causes this issue #262 which lead me to the option to just abort the process once the issue occurs.

I've been running with this for almost 3 days straight already without any unknown packets and restarts so far. I had to restart SquadJS earlier due to a bug fix for the map vote and will continue to test it and run it the whole week.

This is the first time we reached more than 24 hours without restarting and any issues. It's so weird to look at our directory only containing one log file.

ect0s:
Small update above; 

Found that commands run on timers (such as the updatePlayer lists etc) can sometimes run before we have been authorized, added a flag. 

I've let this run for 3 days, including during/over several restarts (including a multihour restart from a server migration). 

So far, no issues.

werewolfboy13 (approved):
Appears to look good, I don't have a server to test on since my RCON is heavily customized.

werewolfboy13:
@fantinodavide can you verify that I pushed this correctly?

steelskillet:
That did not work correctly. Please revert.

fantinodavide:
@werewolfboy13 lots of changes after the merge and even if it works, all the new changes made by ect0s have been removed

werewolfboy13:
Broke the merge, new PR to fix. Closing.

PR #273, opened 2023-01-21
Stage config.json as config.json.example

### LINKED ISSUE #275: Remove "config.json" as a tracked respository file, rename to "config.json.example" to highlight the config file requires manual config.
`git pull`  replaces config.json with the example config by default.
`git pull` should not replace config.json and the provided config.json should be labelled as an example
Renaming config.json and removing the tracking of the file from git allows for a persistant configuration between updates of master without needing to copy/replace the configured config.json file.
Cant see anything in the build / install files as discussed in #273 as been problematic or altering to the config.json file.

See #273.

For a **new** user basing on that commit the following update workflow is functional:
1. git clone https://github.com/Team-Silver-Sphere/SquadJS.git
2. Copy config.json.example to config.json
3. configure config.json
4. install with yarn install
5. (optional) git rm config.json (tells git to ignore the newly created file, possibly mandatory if your git blocks the next pull due to un-commited changes, ie it finds a new file you haven't told it about with git add \<file\> for some reason)
6. run with node index.js

Same steps as previous with the additional step of acknowledging the shipped config file is unconfigured and needs manual intervention.

Once installed the follow workflow can be used to update the install from git without dropping config & run the tool:
1. git pull (no longer replaces config.json because its called called config.json.example upstream)
2. yarn install
3. node index.js

### Additional Info
Add any other info or screenshots about the feature request here.
I've had this in my live branch for a few updates cycles and its functional
Added @ https://github.com/Team-Silver-Sphere/SquadJS/commit/c825cf1241e46c29b612464caa79380d04ecdd3b

Since then i've updated twice (thanks ect0s) and rebased against master, config.json has persisted the entire time.

DrKittens:
Alteration of the config.json filename shipped with the project will allow users to update with `git pull` without overwriting a configured `config.json` file.

Additionally adds a section with "updating" steps to the readme for use for people running from a clone of the repo not the release version.


werewolfboy13 (changes requested):
This will not work, rejecting changes. There is jobs that will overwrite this. Also if you are doing changes templates must be updated.

werewolfboy13:
Look at the build commands.

werewolfboy13:
Will close this out unless you are willing to provide a better system. This change does not have a point as cron jobs will overwrite this. Make a issue ticket if you want first. Will be willing to reconsider this at a future time. 

This change will require alot more changes to work.

PR #274, opened 2023-01-21
Improve tank detection in layer data

fantinodavide:


werewolfboy13:
So I am the pipeline maintainer, what is the use case again for this?

fantinodavide:
The regex `/tank/` is used to determine if a vehicle is a tank, but was also matching other vehicles whose icon is "map_jeep_antitank"

werewolfboy13:
Have we checked that no tanks will be missed with this change? I can check the SDK output if needed.

image

ect0s (approved):
This Looks correct to me, 

I did pull down the layers.json and run a few grep expressions on it. Without this fix, we do collide on jeep_antitank. 

And with the fix it appears to be correct. 

The only edge case, would be if the "icon" type is mislabeled on a vehicle which would be upstream in the wiki code. 
```
cat _tank.txt | sort | uniq -c
    415             "icon": "map_tank",
cat tank.txt | sort | uniq -c
    194             "icon": "map_jeep_antitank",
    415             "icon": "map_tank",
```

werewolfboy13 (approved):

PR #277, opened 2023-01-24
Expose a2s playtime on server

reck1610:
This exposes the a2s playtime information on the SquadJS server and allows it to be used in plugins.

### DIFF
--- a/squad-server/index.js
+++ b/squad-server/index.js
@@ -457,6 +457,7 @@ export default class SquadServer extends EventEmitter {
         publicQueue: parseInt(data.raw.rules.PublicQueue_i),
         reserveQueue: parseInt(data.raw.rules.ReservedQueue_i),
 
+        playtimeSeconds: parseFloat(data.raw.rules.PLAYTIME_i),
         matchTimeout: parseFloat(data.raw.rules.MatchTimeout_f),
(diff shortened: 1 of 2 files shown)


Thomas-Smyth:
Does this reflect a change in Squad? Previously the match timeout counted downwards, but your changes seem to imply it doesn't now and that data is replaced by this new value?

ect0s:
matchTimeout appears to stay at the 120 minutes? (2 hours?)

```"MatchTimeout_f":"120.000000"```

playtime_i appears to increment. 

```"PLAYTIME_i":"1889"```

Thomas-Smyth:
Perhaps we should consider putting these in the same format, I.e. both in seconds or minutes?

watchful-eyes:
`Playtime_i` data increases even if the server is on waiting mode(has no players) like it can be 6 hours if noone join, so the data partially not reliable I guess.

Thomas-Smyth:
> `Playtime_i` data increases even if the server is on waiting mode(has no players) like it can be 6 hours if noone join, so the data partially not reliable I guess.

Got to love Squad...

werewolfboy13:
Closing this due to changes coming down the pipeline soon.

PR #281, opened 2023-02-05
Improve custom plugins

### LINKED ISSUE #280: Improve custom plugin support
In my opinion, custom plugins have 2 problems at the moment.

#### 1. Custom base classes for plugins will always be loaded as a plugin.
This might not cause a problem, but it causes custom plugins to be handled differently from default plugins.
Currently, the default base classes are blacklisted in the plugin loader. 

#### 2. Custom plugins have to be placed in a hard-coded location in the repo
This causes git changes and makes updating SquadJS harder.
This also reduces maintainability for both custom and default plugins, if custom plugins are used. 

**Problem1:** I would like to introduce the file extension `.squad.js` for plugins. This removes the need for blacklisting other files in the plugin directory.

**Problem 2:** I would like to make it configurable from where SquadJS is loading the plugins. Furthermore, I would like to allow multiple paths to be configured in order to increase maintainability for both custom and default plugins, if custom plugins are used. 
I would like to introduce `pluginPaths` into the `server` configuration like this:
``` json
{
  "server": {
    "id": 1,
    "host": "xxx.xxx.xxx.xxx",
    "queryPort": 27165,
    "rconPort": 21114,
    "rconPassword": "password",
    "logReaderMode": "tail",
    "logDir": "C:/path/to/squad/log/folder",
    "ftp": {
      "port": 21,
      "user": "FTP Username",
      "password": "FTP Password",
      "useListForSize": false
    },
    "pluginPaths": [
      "squad-server/plugins/",
      "/opt/custom-squadjs-plugins/"
    ]
  }
```

Both relative and absolute paths should be allowed. Plugins will be loaded in order (starting with index 0), if a plugin already exists it will be replaced. This should be logged accordingly.

### Additional Info
I am already working on this.
This relates to #275.

reck1610:
The changes I proposed in #280.

closes #280


werewolfboy13:
This would require major rewrites on the documentation and best internal practices. This request is at a level of a major rewrite and will massively break a ton of other PRs.

werewolfboy13:
At this time this is not something I want to push through.

PR #283, opened 2023-02-05
Improve configuration management for multiple servers

### LINKED ISSUE #275: Remove "config.json" as a tracked respository file, rename to "config.json.example" to highlight the config file requires manual config.

### LINKED ISSUE #282: Improve configuration management for multiple servers
It is tedious to manage configurations for multiple squad servers and keep them in sync.

I would like SquadJS to allow multiple configuration to be loaded.
The configurations should be passed as additional parameters like this:
``` shell
node index.js config.base.json config.server1.json
```

The configurations should be deep merged. Changes in later configurations should overwrite earlier configurations. 
Arrays should be concatenated object properties should be replaced.

reck1610:
This allows you to split up your configuration and reuse parts on multiple servers.

This changes the example configuration name as proposed in #275 

closes #282, closes  #275


.gitignore
```diff
@@ -12,3 +12,7 @@ yarn.lock
 # IDEs
 .idea/
 .vs/
+
+config.json
+config.*.json
+!config.example.json
```
werewolfboy13, line 18: Why is this excluded?

werewolfboy13 (changes requested):
This will need vetting with multiple live servers before it is approved. Also need an answer on the code comment.

.gitignore
reck1610: `config.json` and `config.*.json` exclude customized configurations.

`config.example.json` is the auto generated configuration example (formerly `config.json`) and explicitly included because it matches `config.*.json`.

steelskillet:
Rather than defining what configs to layer when you run squadjs. I think it would be better to have an include directive (eg `!include "path to other config"`). This would work better for automation systems that use squadjs and would keep the ordering and what other configs are required for a particular configuration inside that configuration file.

werewolfboy13:
Review stalled moving this over to #320 

**CLOSING DUE TO LACK OF RESPONSE**

PR #286, opened 2023-02-12
Bump Version Number

werewolfboy13:
Updates Package JSON version


werewolfboy13:
Bumping to v3.7.0. Dont know the flag system so @Thomas-Smyth would need to add flags as needed.

PR #314, opened 2023-11-02
Improve Layer finding

steelskillet:
Goal: Currently the RCON overrides any layer found by the log parser. RCON fails to find the correct layer more often and provides not up to update information on the current layer. These changes prioritize Layer updating in order of accuracy and adds using the layer information from the the A2S query (which gives layer id as opposed to RCON which gives layer name).

Idea: Basic idea of the logic is that if the current layer in `this.currentLayer` gives the same information that is given from the squad server (for either A2S or RCON) then it should not update the current layer to whatever it would possibly find (either no layer or an incorrect layer).

Special Considerations: This code also takes into account a race condition with the log parser in which if we start the RCON/A2S query then compare to the current layer data, it's possible that the log parser changed the current layer data after the squad server gave the information to the query but before that query returned in squadjs. Does this by pulling the current layer data before initiating the query.

Simpler logic could not be used (eg `if (this.currentLayer == null) update layer`) in the rare case that we lose the log parser (eg the log parser is on ftp or similar).

Testing: Currently not tested and needs sanity checking (I use different logic for my squadJS at the moment, will be copying this in and testing shortly).

### DIFF
--- a/squad-server/index.js
+++ b/squad-server/index.js
@@ -412,11 +412,18 @@ export default class SquadServer extends EventEmitter {
     Logger.verbose('SquadServer', 1, `Updating layer information...`);
 
     try {
+      let currentLayer = this.currentLayer;
       const currentMap = await this.rcon.getCurrentMap();
       const nextMap = await this.rcon.getNextMap();
       const nextMapToBeVoted = nextMap.layer === 'To be voted';
 
-      const currentLayer = await Layers.getLayerByName(currentMap.layer);
+      if (currentLayer?.name !== currentMap.layer){
+        const rconlayer = await Layers.getLayerByName(currentMap.layer);
+        if (rconlayer){
+          currentLayer = rconlayer;
+        }
+      }
+      
       const nextLayer = nextMapToBeVoted ? null : await Layers.getLayerByName(nextMap.layer);
 
       if (this.layerHistory.length === 0) {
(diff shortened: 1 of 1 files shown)


werewolfboy13:
Recommended here is RCON first as it is more reliable and will have to be the focus for changes to Squad that are planned down the road. I'd rather keep it RCON first if that is the current method.

steelskillet:
Current method is if either the `NEW_GAME` or `updateLayerInformation` events run they unconditionally set `this.currentLayer` to whatever they think the layer is or isn't, it just so happens that the RCON based updateLayerInformation gets called periodically and thus destroys any information that the log parser may have found, even when the RCON doesn't find the layer it overwrites it with null. The log parser has a higher chance of finding the correct layer because it uses the layer id rather than the layer name.
With this as well, the current information is essentially double checked by RCON and if found incorrect then RCON will update it if it finds 'better' information. 
A2S works the same way and is given equal priority with regards to RCON however has a higher chance of finding a correct layer due to reporting layer id rather than layer name (it also required minimal changes).
It's more along the lines of if one of the queries finds bad information then it shouldn't overwrite what's already there.

werewolfboy13:
> Testing: Currently not tested and needs sanity checking (I use different logic for my squadJS at the moment, will be copying this in and testing shortly).

This will need public testing before I can push anything through.

steelskillet:
testing started for ASG servers.

PR #315, opened 2023-11-03
Improvement of the default packages

fantinodavide:
The aim of this PR is to include some useful packages to facilitate sharing custom plugins that require more complex processing.

Canvas, paired with Chart.js can be used to generate graphs and to render them to images.

Tar, can be used to create a tarball of multiple files, and eventually to compress the output using the gzip algorithm, making easier to share groups of files, for example backups.

### DIFF
--- a/squad-server/package.json
+++ b/squad-server/package.json
@@ -4,6 +4,8 @@
   "type": "module",
   "dependencies": {
     "axios": "^0.21.1",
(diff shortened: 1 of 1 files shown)


Thomas-Smyth:
There's no reason to add these packages if something is not actively using them. If a pull request is utilised containing a plugin that utilises them, only then will we consider adding them.

PR #320, opened 2023-11-12
Multi-config using config includes

steelskillet:
Based on changes in #283.
Provides specifying config files to include before and after a config file to append to a config list. It then applies the values from each json config in the list in order overwriting previous values. includes are recursive and errors on circular dependency.

This allows easily sharing a bundle of configs and puts into the actual configuration, what other configs are required for a specific config file.

needs extended testing (changes will be merged into ASG servers), and a double check of the changes to the documentation at the moment (and maybe a little clean up of the code).


steelskillet:
small test with ASG servers, seems to be working as expected. 
Worked with no change to original config file, both includes and overwrites seems to be working as expected as well.

I would like to get a better log entry for printing the config out, one that lists the absolute path of the file that put the value in for each config entry to help out with debugging what configs are setting what value. This will require some non-trivial changes however and might take a bit.

werewolfboy13:
Needs rewrite

steelskillet:
if it is just the conflicts, we good now. if there is anything specific then list would be good.

steelskillet:
This doesn't work how i imagined it in the current state, i plan on working on it fairly soon (tm) so i will leave this open for now, just expect large changes or this getting closed and a new one reopened.

Issue #289, opened 2023-03-04
'PLAYER_DISCONNECTED' and RCON race condition results in null player


ghost:
### Description of Issue
A 'PLAYER_DISCONNECTED' event can pass a null player due to a race condition with updating the RCON player list.
If the the main squadjs player list is updated via RCON before the log parser is done updating, a player that causes a disconnect event through the log parser will not exist in the main player list because the RCON updated it before the log parser caught the disconnect event.

### Errors or Screenshots of Issue
This happened a while ago (and i created a workaround in my fork of the repo) and thus i don't have logs for it atm. Some suggested fixes would be to manually fill out the player information from the raw information provided by the log parser (essentially what i did for workaround), or to keep track of players that have also recently disconnected. 

Log essentially just had the RCON update with a player just moved to the disconnect section then directly after, debug messages for saying that a player disconnect event had happened and then a debug message for saying that it couldn't find the player in the main player list.

### Squad Information
 * SquadJS Version: main branch
 * Squad Version: v4.2

### Squad Server State
Full player load with lots of log parsing being required.

### System Information
 * Operating System: linux
 * NodeJS Version: n/a
 * Yarn Version: n/a

PR #321, opened 2023-11-12
Bug fix for 'PLAYER_DISCONNECTED' race condition

steelskillet:
Fixes the bug described in #289.

Implements this by saving the list of `oldPlayerInfo` that `updatePlayerList()` creates and then the `PLAYER_DISCONNECT` event tries to find the player by steam id in the saved `oldPlayerInfo` list if it couldn't find it in `this.players`.


steelskillet:
Started testing on ASG servers.

werewolfboy13:
Needs a rewrite for the new major squad js version

steelskillet:
updated, needs further testing still.

PR #322, opened 2023-11-18
Fix typo in auto-kick-unassigned.js

vohk:
Fixed typo in unassigned warning; "you are are unassigned"


werewolfboy13:
Will need a config rerun to make it stick at some point

PR #323, opened 2023-11-18
Add includeChats property

CodeRedDev:
This adds a `includeChats` property to `ChatCommands`.

This makes limiting commands to one specific channel much easier and cleaner in the configuration.

I saw that there are more modules with the `ignoreChats` property and if the new `includeChats` property is desired, I will make efforts to implement it for those other modules too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,14 +295,15 @@ Interested in creating your own plugin? [See more here](./squad-server/plugins/r
           <h3>Options</h3>
           <ul><li><h4>commands</h4>
            <h6>Description</h6>
-           <p>An array of objects containing the following properties: <ul><li><code>command</code> - The command that initiates the message.</li><li><code>type</code> - Either <code>warn</code> or <code>broadcast</code>.</li><li><code>response</code> - The message to respond with.</li><li><code>ignoreChats</code> - A list of chats to ignore the commands in. Use this to limit it to admins.</li></ul></p>
+           <p>An array of objects containing the following properties: <ul><li><code>command</code> - The command that initiates the message.</li><li><code>type</code> - Either <code>warn</code> or <code>broadcast</code>.</li><li><code>response</code> - The message to respond with.</li><li><code>ignoreChats/includeChats</code> - Either one can be used to limit to specific in-game chats, being <code>ChatAll</code>, <code>ChatTeam</code>, <code>ChatSquad</code> and <code>ChatAdmin</code></li></ul></p>
            <h6>Default</h6>
            <pre><code>[
   {
     "command": "squadjs",
     "type": "warn",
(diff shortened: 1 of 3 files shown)


lbzepoqo:
I was about to do another PR which might be in conflict to yours, maybe you can add it here too instead?

## Add the ability for "commands" option to accept an array or just a single string.
```json
  {
      "command": [
          "claim",
          "claimrule"
      ],
      "type": "broadcast",
      "response": "To make a valid claim, name your squad after the vehicle you want to claim.\nIf there are multiple vehicles of the same type, the first squad to claim it gets priority.",
      "ignoreChats": [
          "ChatAll",
          "ChatTeam",
          "ChatSquad"
      ]
  },
```
```js
  async mount() {
    for (const command of this.options.commands) {
      const commands = Array.isArray(command.command) ? command.command : [command.command];
      for (const cmd of commands) {
        this.server.on(`CHAT_COMMAND:${cmd.toLowerCase()}`, async (data) => {
          if (command.ignoreChats.includes(data.chat)) return;
  
          if (command.type === 'broadcast') {
            await this.server.rcon.broadcast(command.response);
          } else if (command.type === 'warn') {
            await this.server.rcon.warn(data.player.steamID, command.response);
          }
        });
      }
    }
  }
  ```

**Comment by CodeRedDev** [comment]
@sergelouie6 although it's a nice feature I doubt that they should be mixed in one PR

File: README.md [hunk 1]
```diff
@@ -295,14 +295,15 @@ Interested in creating your own plugin? [See more here](./squad-server/plugins/r
           <h3>Options</h3>
           <ul><li><h4>commands</h4>
            <h6>Description</h6>
-           <p>An array of objects containing the following properties: <ul><li><code>command</code> - The command that initiates the message.</li><li><code>type</code> - Either <code>warn</code> or <code>broadcast</code>.</li><li><code>response</code> - The message to respond with.</li><li><code>ignoreChats</code> - A list of chats to ignore the commands in. Use this to limit it to admins.</li></ul></p>
+           <p>An array of objects containing the following properties: <ul><li><code>command</code> - The command that initiates the message.</li><li><code>type</code> - Either <code>warn</code> or <code>broadcast</code>.</li><li><code>response</code> - The message to respond with.</li><li><code>ignoreChats/includeChats</code> - Either one can be used to limit to specific in-game chats, being <code>ChatAll</code>, <code>ChatTeam</code>, <code>ChatSquad</code> and <code>ChatAdmin</code></li></ul></p>
```
Thomas-Smyth, line 298: Were these changes made manually? If so, they will be overwritten when the readme is regenerated from the plugin option specification.

Thomas-Smyth (changes requested):

PR #327, opened 2023-11-21
Added timestamps option to logger configuration

lbzepoqo:
This is in line with https://github.com/Team-Silver-Sphere/SquadJS/pull/271

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,8 @@ The `logger` section configures how verbose a module of SquadJS will be as well 
         "SquadServerFactory": "yellowBright",
         "LogParser": "blueBright",
         "RCON": "redBright"
-      }
+      },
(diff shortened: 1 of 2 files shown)


Thomas-Smyth:
If we think that timestamps should be off by default - I am open to that debate - it should be made the default through the code rather than the config files. In either case, this pull request does not change the template files, and thus it will be overwritten by husky whenever someone commits (it appears husky was not run when you made these commits). Given there does not appear to be much support for changing the default behaviour on this pull request and the issues with how the change is implemented, I am going to close this pull request. If you think timestamps should be off by default, please implement it as default through the code and then open a new pull request. However, that will only be accepted if there is a wider agreement that it should be the default.

PR #341, opened 2024-01-06
Rollback on commit

### LINKED ISSUE #338: JS file check breaks application
The introduced JS file check from #319 breaks the application. Non of the plugins will be inserted into the server plugins list and will not be mounted at all.

CodeRedDev:
Removes the unnecessary and bug triggering check for .js postfix on `plugin` property from config

Fixes #338

### DIFF
--- a/squad-server/factory.js
+++ b/squad-server/factory.js
@@ -81,11 +81,6 @@ export default class SquadServerFactory {
     for (const pluginConfig of config.plugins) {
       if (!pluginConfig.enabled) continue;
 
-      // ignore non JS files
-      if (!pluginConfig.plugin.endsWith('.js')) {
-        continue;
-      }
-
       if (!plugins[pluginConfig.plugin])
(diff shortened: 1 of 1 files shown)


werewolfboy13:
Not approving PR unless supporting documentation is given.

CodeRedDev:
@Zer0-1ne put the check this PR is removing here by accident and unfortunately did not remove them in his latest commits.

The check here inside `factory.js` does not check against file identifiers it check against the `plugin` properties of the plugins inside `config.json` and as they are never having a `.js` postfix, non of the plugins will be added to the servers plugins list and non will be mounted.

werewolfboy13:
Apologies, I see what is going here. My IDE was not loading this PR correctly. Pushing for approval.

werewolfboy13:
@Zer0-1ne I will need you to double check your other commits to make sure its placed correctly when able.

Zer0-1ne:
I seriously dont understand why it worked beforehand, but I've checked and its in the correct place, sorry about that.

PR #344, opened 2024-01-24
Feat: Add @here & Server Name

IgnisAlienus:
Adds the ability to turn on @here for Admin Requests in Discord. This allows for pinging of all ONLINE Squad Admins if the Admin Request is going to a Squad Admin only Channel.

Added the ServerName option so you can tell what server the request is coming from for communities with multi servers without having to create multi channels for each server's admin request notifs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,11 +447,6 @@ Grafana:
            <p>Should players know how much in-game admins there are active/online?</p>
            <h6>Default</h6>
            <pre><code>true</code></pre></li>
-<li><h4>serverName</h4>
-           <h6>Description</h6>
-           <p>The Server Name will come across in the Discord Message</p>
-           <h6>Default</h6>
-           <pre><code>Server 1</code></pre></li></ul>
         </details>
 
 <details>
--- a/README.md
+++ b/README.md
@@ -422,6 +422,11 @@ Grafana:
            <pre><code>[
   "500455137626554379"
 ]</code></pre>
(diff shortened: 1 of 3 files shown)


IgnisAlienus:
Updated to use a this.server.serverName instead of using a plugin option of this.options.serverName.

fantinodavide:
The readme file hasn't been updated when removing the server name option, also, updating it shouldn't be a manual process as there is a command and a pre-commit hook to automatically update the config file, readme, and lint/format the files affected by the changes.

IgnisAlienus:
> The readme file hasn't been updated when removing the server name option, also, updating it shouldn't be a manual process as there is a command and a pre-commit hook to automatically update the config file, readme, and lint/format the files affected by the changes.

Updated ReadME. 

Yeah, things are wonky in my fork/branch so it was easier for me to just modify the files directly in this PR.

werewolfboy13 (changes requested):
Please do not manually edit the readme, please use `yarn lint`.

Please push a commit with that edit done if it hasn't already been ran.

werewolfboy13:
@IgnisAlienus I see you removed the server name, was that intended?

IgnisAlienus:
> @IgnisAlienus I see you removed the server name, was that intended?

Yes. this.server.serverName returns the Server Name that I is fetched somewhere else in SquadJS

werewolfboy13 (approved):

PR #345, opened 2024-02-01
Feat: Support for GC's Player Controllers

IgnisAlienus:
Support for GC's Player Controllers.

I added them in the log-parser as new files instead of messing with and possibly breaking old files.

This may need to be done for any other mods that mess with the BP name for the Player Controller.

Let me know if there's a better way to implement this. But this is working for me on our Server.

### DIFF
--- a/squad-server/log-parser/index.js
+++ b/squad-server/log-parser/index.js
@@ -4,7 +4,9 @@ import AdminBroadcast from './admin-broadcast.js';
 import DeployableDamaged from './deployable-damaged.js';
 import NewGame from './new-game.js';
 import PlayerConnected from './player-connected.js';
+import PlayerConnectedGC from './player-connected-gc.js';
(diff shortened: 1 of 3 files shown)


steelskillet:
might be able to just loosen the regex for the player controllers. I don't think they need the current "BP_PlayerContoller_C" and we might just be able to use ".+" instead.

IgnisAlienus:
> might be able to just loosen the regex for the player controllers. I don't think they need the current "BP_PlayerContoller_C" and we might just be able to use ".+" instead.

Yeah, that would work just as well if it won't break anything else.

werewolfboy13:
Is there plugins out now that would benefit from this?

IgnisAlienus:
People that have custom time tracking plugins. I think the other guy my be onto something where the regex on the one is made less specific for the controller. As long as that doesn't break anything else.

Issue #299, opened 2023-07-15
Allow FTP Address to be specified


Oracuda:
Some hosts have a different address FTP is to be accessed to other than the server address. please make this configurable in config.json.

Quick workaround is to hardcode the FTP address.

PR #346, opened 2024-02-12
Added host property to ftp config to limit confusion as this is a common configurable property.

mbartisan:


CodeRedDev:
This might require a `ftp:host` description in the `README.md` to be complete.

Even though I'm questioning the sense in this parameter overall. If I'm not mistaken then the log should always be on the server that hosts the game server. So there should not be a parameter to set another `ftp:host` and this line

https://github.com/Team-Silver-Sphere/SquadJS/blob/923c601292cf9264fd8b30cfa7baec9acefb45d8/squad-server/index.js#L184 

should always take `this.options.host`.

mbartisan:
Specifying the ftp host is required to work with some server providers. See Issue #299 where this feature was implemented. This PR is adding documentation to the ability.

CodeRedDev:
Ah alright sorry.

My first point should be valid though that the `ftp:host` should be descibed here

https://github.com/Team-Silver-Sphere/SquadJS/blob/afcab47e8122b796cc45b1a7788e9a2013b837f0/README.md?plain=1#L94

Maybe also by mentioning #299 👍

PR #347, opened 2024-02-23
Update admin list utility to recognise eosIDs

IgnisAlienus:
So, I updated my admins.cfg to have EOSIDs for my Admins after that whole EOS Update, guess SquadJS hasn't been pulling them this whole time. Now it does.

### DIFF
--- a/squad-server/utils/admin-lists.js
+++ b/squad-server/utils/admin-lists.js
@@ -44,7 +44,7 @@ export default async function fetchAdminLists(adminLists) {
     }
 
     const groupRgx = /(?<=^Group=)(?<groupID>.*?):(?<groupPerms>.*?)(?=(?:\r\n|\r|\n|\s+\/\/))/gm;
-    const adminRgx = /(?<=^Admin=)(?<steamID>\d+):(?<groupID>\S+)/gm;
+    const adminRgx = /(?<=^Admin=)(?<adminID>\d{17}|[a-f0-9]{32}):(?<groupID>\S+)/gm;
 
     for (const m of data.matchAll(groupRgx)) {
       groups[`${idx}-${m.groups.groupID}`] = m.groups.groupPerms.split(',');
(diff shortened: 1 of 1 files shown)


fantinodavide (approved):
Changes look good to me. Approved

Thomas-Smyth:
How does this work when utilising the list? A lot of the Server class seems to search the list of admins by Steam IDs still, e.g.:
https://github.com/Team-Silver-Sphere/SquadJS/blob/9f9755ed059de608389229e1b72b3425e8b22e4c/squad-server/index.js#L357

Ulibos:
@Thomas-Smyth I'm building a full migration to eosIDs. I'm planning to implement a mapping function that will convert steamIDs of online admins into eosIDs (non-matching IDs will be ignored instead of failing). This will allow us to support old steamID-based admin lists while people shift their stuff to eosIDs. My future PR builds upon this PR and my own: https://github.com/Team-Silver-Sphere/SquadJS/pull/352

fantinodavide:
> How does this work when utilising the list? A lot of the Server class seems to search the list of admins by Steam IDs still, e.g.:
> 
> https://github.com/Team-Silver-Sphere/SquadJS/blob/9f9755ed059de608389229e1b72b3425e8b22e4c/squad-server/index.js#L357

it accepts both steamID and eosID

Thomas-Smyth:
> > How does this work when utilising the list? A lot of the Server class seems to search the list of admins by Steam IDs still, e.g.:
> > https://github.com/Team-Silver-Sphere/SquadJS/blob/9f9755ed059de608389229e1b72b3425e8b22e4c/squad-server/index.js#L357
> 
> it accepts both steamID and eosID

How? If it's always looking up by Steam ID then it will never match on a EOSID.

fantinodavide:
> > > How does this work when utilising the list? A lot of the Server class seems to search the list of admins by Steam IDs still, e.g.:
> > > https://github.com/Team-Silver-Sphere/SquadJS/blob/9f9755ed059de608389229e1b72b3425e8b22e4c/squad-server/index.js#L357
> > 
> > 
> > it accepts both steamID and eosID
> 
> How? If it's always looking up by Steam ID then it will never match on a EOSID.

People will have to adjust their plugins in order to use eosID, but this PR improves how the admin lists are parsed and won't skip the lines with an eosID only, and doesn't make a specific discrimination between steamID and eosID

Ulibos:
> How? If it's always looking up by Steam ID then it will never match on a EOSID.

If an ID is used directly in rcon commands then it will work regardless. There are 2 plugins that will ignore eosIDs because they loop through players matching admin IDs against steamID:
1. `auto-kick-unassigned.js`
2. `discord-admin-request.js`

But I'm planning to address this in my next PR.

werewolfboy13:
Pending further review.

PR #349, opened 2024-03-09
Update CBL plugin: allow flexible kicking of untrusted players

RobotoLev:
Wrote this code for myself to be able to kick known cheaters automatically. Now proposing it to everyone to use.

Full list of ban-filtering abilities:
1. By length ("not less days than")
2. By ban tags ("if any of these is present")
3. By ban issuer ("if any of these is present")

All of them are optional.

This PR also adds steamID to Discord embed messages of the plugin in order to simplify moderation processes.


RobotoLev:
Example of the plugin in work (edited for privacy):
cbl_example

PR #352, opened 2024-03-26
Robust ID parser.

Ulibos:
This patch resolves an issue with inconsistent ID reports from RCON and game logs. An example to illustrate the issue:

```
----- Active Players -----
ID: 1 | Online IDs: EOS: 00023b618e9c4e6e8e85122d664bdbd9 steam: 76561198923824046 | Name: [DTX] AlreadyGod | Team ID: 2 | Squad ID: 1 | Is Leader: True | Role: PLA_Medic_01
ID: 0 | Online IDs: EOS: 0002eda25e7543f18f327956dac87509 | Name:  Ulibos | Team ID: 2 | Squad ID: 2 | Is Leader: True | Role: PLA_Medic_01
----- Recently Disconnected Players [Max of 15] -----
```

Note that my profile is missing a steam ID. This is not a synthetic example, this actually happened to me and happens sometimes to others. This leads to a bunch of parsers failing to match the line and simply discarding it making the player non-existent for squadjs.

**What it does:**
Parsers read all IDs in bulk and defer matching of individual IDs to a separate function. This should also protect against potential future introduction of new platforms and IDs.

**What it doesn't:**
This PR does not fix the fact of absence of a steam ID of a player. If some functionality still relies on that, it will fail. If there is no check for an undefined steam ID somewhere, it might lead to failures. I'm hoping to address this in my next PR.

### DIFF
--- a/core/id-parser.js
+++ b/core/id-parser.js
@@ -1,0 +1,41 @@
+const ID_MATCHER = /\s*(?<name>[^:]+): (?<id>[^\s]+)/g;
+
+// PARSING AND ITERATION
+
+// main function intended for parsing `Online IDs:` body. Returns an
+// iterator that yields `{platform: id}` pairs.
+export const iterate = (idsStr) => {
+  return new IdsIterator(idsStr.matchAll(ID_MATCHER));
+};
+
+class IdsIterator {
+  constructor(matchIterator) {
+    this.inner = matchIterator;
+  }
+
+  [Symbol.iterator]() {
+    return this;
+  }
+
+  next() {
+    const match = this.inner.next();
+    if (match.done) return { value: undefined, done: true };
+    return { value: { key: match.value[1], value: match.value[2] }, done: false };
+  }
+
+  forEach(callbackFn) {
+    for (const { key, value } of this) callbackFn(key, value);
+  }
+}
+
+// FORMATTING
+
+// `steam => SteamID`
+export const capitalID = (str) => {
+  return str.charAt(0).toUpperCase() + str.slice(1) + 'ID';
+};
+
+// `EOS => eosID`
+export const lowerID = (str) => {
+  return str.toLowerCase() + 'ID';
+};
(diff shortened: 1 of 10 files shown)


lbzepoqo:
Sample log lines of player connected without steam ids. Explains why both of us lost admin perms this day

```
[2024.02.16-06.28.21:920][471]LogSquad: PostLogin: NewPlayer: BP_PlayerController_C /Game/Maps/Sumari/Gameplay_Layers/Sumari_AAS_v3.Sumari_AAS_v3:PersistentLevel.BP_PlayerController_C_2147481999 (IP: 120.29.86.10 | Online IDs: EOS: 00020661cc07480cb8bd10b3c4bf9104)
[2024.02.16-06.29.01:120][228]LogSquad: PostLogin: NewPlayer: BP_PlayerController_C /Game/Maps/Sumari/Gameplay_Layers/Sumari_AAS_v3.Sumari_AAS_v3:PersistentLevel.BP_PlayerController_C_2147481855 (IP: 158.62.51.10 | Online IDs: EOS: 00029e5d00e947bbaed1b364814dd427)
```

werewolfboy13:
Concerned with the potential breaking of other plugins due to regex being redone.

Ulibos:
> Concerned with the potential breaking of other plugins due to regex being redone.

It shouldn't brake anything by itself. What can happen is it will parse a line without steamID instead of discarding it (because now regex is adapted to dynamic number of IDs) and something down the line will die because trying to use an `undefined` in calculations or indexing a player with `undefined` (and getting an `undefined` in return) and trying to access it's fields. But that is basically addressed in #354
The new regex itself is pretty trivial. `Online IDs` section is always delimited by either `()` or `[]` or `||`. Each regex simply grabs an entire chunk till the closing delimiter and passes it to a looped ID parser. You can give it 1,2, 200 ids and as long as they maintain the same format (delimiters, `:` and ` ` are reserved tokens) - it will crunch through it.

Ulibos:
Closing this PR because it was embedded in #354

PR #354, opened 2024-04-15
Improve migration to eosIDs

Ulibos:
# Summary

This PR completes migration to eosIDs for the tool (except for a couple plugins, more on that below).

## Dependencies

This PR depends on #352 and lays foundation for #347 

## Changes

There are 5 commits:
1. Cleanup player connection log parsing to ease transition and get rid of clutter.
2. Rework and extend admin-related methods to make them "omnivore" to different IDs. *
3. Migrate core code to eosIDs.
4. Migrate plugins to eosIDs. *
5. Fix minor issue with uninitialized `oldPlayerInfo` on player list updates (was annoying me a bit during testing).

_* - caveats_

## Testing

All of this was tested several times with event capture to make sure data is consistent. In summary (but not in 1 last go) I have tested:

**connecting**, **disconnecting**, **possess**, **unpossess**, **create squad**, **kill** (self inflicted), **warn**, **force swap**, **kick**, **ban**.

Only thing I didn't test for sure is a **teamkill**. I also am not aware if I had at least 1 run where I didn't have a steamID, most likely not. But so far generalized parsers work correctly, I haven't seen any errors in logs or missing data. I have also tested: `getAdminPermsByAnyID()`, `getAdminsWithPermission()` (both were tested for `server.admins` with steamID or eosID), `getPlayerByAnyID()` for both types of IDs, rcon methods (`warn()`, `switchTeam()`, `kick()`, `ban()`).

### Testing Plugins
I didn't test plugins. We don't use any default ones and we don't have the necessary setup, plus the ones I've edited have mostly trivial fixes. The one I've fixed only partially is **discord-killfeed**, I didn't touch it's CBL integration. The ones I didn't touch at all are: **db-log**, **awn-api**, **cbl-info**. Some of these are completely unknown to me, some are too dependent on steamID to fix them in this PR. For now they will work as they were before as long as players have steamIDs.

### Admins IDs
For now, there is a potential issue with an admin having 2 sets of permissions, one for eosID and another for steamID. Current code won't be able to merge it (at the time of merging it doesn't have persistent mapping of player's IDs). Easily avoidable though by not intermixing two ID formats.
Another side-effect is that `getAdminsWithPermission()` will always return only those admins that have the permission AND are online (for at least 10 seconds). This happens because this method matches all existing IDs against all existing admin entries to convert them all to eosIDs.


squad-server/index.js
```diff
@@ -353,16 +336,26 @@ export default class SquadServer extends EventEmitter {
     await this.logParser.watch();
   }
 
+  // DEPRECATED FUNCTION
```
fantinodavide: can be replaced with a more suitable jsdoc
```
/**
 * @deprecated <optional message here>
 */
 ```

File: core/id-parser.js
Comment (fantinodavide) [inline core/id-parser.js:1]: I personally don't like the approach of having a "unified" regex to match the IDs because it depends only on the hope of OWI not breaking things with typos or non-standard formats, which has already happened, an example is a log line which has `Contoller` instead of `Controller`.

File: squad-server/index.js [hunk 2]
```diff
@@ -218,7 +218,7 @@ export default class SquadServer extends EventEmitter {
       data.player = await this.getPlayerByEOSID(data.eosID);
       if (data.player) data.player.suffix = data.playerSuffix;
 
-      delete data.steamID;
+      for (const k in data) if (k.endsWith('ID')) delete data[k];
```
fantinodavide: this should be more specific:
```js
for (const k in data) if (["steamID", "eosID"].includes(k)) delete data[k];
```

squad-server/index.js
```diff
@@ -227,7 +227,7 @@ export default class SquadServer extends EventEmitter {
     this.logParser.on('PLAYER_DISCONNECTED', async (data) => {
       data.player = await this.getPlayerByEOSID(data.eosID);
 
-      delete data.steamID;
+      for (const k in data) if (k.endsWith('ID')) delete data[k];
```
fantinodavide: this should be more specific:
```js
for (const k in data) if (["steamID", "eosID"].includes(k)) delete data[k];
```

squad-server/index.js
```diff
@@ -353,16 +336,26 @@ export default class SquadServer extends EventEmitter {
...
 
   getAdminsWithPermission(perm) {
     const ret = [];
-    for (const [steamID, perms] of Object.entries(this.admins)) {
-      if (perm in perms) ret.push(steamID);
+    for (const [anyID, perms] of Object.entries(this.admins)) {
+      if (perm in perms) ret.push(anyID);
     }
-    return ret;
+    return anyIDsToPlayers(ret, this.players).map(player => player.eosID);
```
fantinodavide: considering that this is a breaking change, and it also limits the functionality from getting ALL the admins to ONLY THE ONLINE admins, I would prefer this to be converted to returning an array of objects with both eosID and steamID optional parameters.
this should be further discussed.

squad-server/index.js
```diff
@@ -591,6 +585,7 @@ export default class SquadServer extends EventEmitter {
     );
   }
 
+  // DEPRECATED
```
fantinodavide: can be replaced with a more suitable jsdoc
```
/**
 * @deprecated <optional message here>
 */
 ```

File: core/id-parser.js [hunk 6]
```diff
@@ -0,0 +1,41 @@
+const ID_MATCHER = /\s*(?<name>[^:]+): (?<id>[^\s]+)/g;
+
+// PARSING AND ITERATION
+
+// main function intended for parsing `Online IDs:` body. Returns an
+// iterator that yields `{platform: id}` pairs.
+export const iterate = (idsStr) => {
```
fantinodavide: this should be renamed to a more descriptive name such as `iterateIDs`

fantinodavide (changes requested):
Looking good overall, but some changes are required in my opinion.

fantinodavide:
Additionally, the "Fix/" prefix is not appropriate in my opinion, and I would change it to "refactor!: " (see https://www.conventionalcommits.org/en/v1.0.0/)

werewolfboy13:
Flags added as needed. Let me know when good to push.

core/id-parser.js
Ulibos: All suggestions are implemented apart from one: changing `getAdminsWithPermission(perm)` to return players instead of IDs is not yet implemented. I'm all for implementing it, just want to make sure it is greenlit. Or, as a compromise, we could extend it to `getAdminsWithPermission(perm, returnPlayers=false)`, that would allow for the new feature without potentially breaking custom plugins.
As for unified regex, I kindly disagree. `master` implementation is prone to: changing ordering, extra spacing, lack of steamID (some regexes in `master` were fixed to have steam as an optional but not all), renaming a key. This implementation will break only if delimiters are broken. Swapping, cloning, removing arbitrary k:v pairs, introducing new IDs, surrounding colons with spaces won't break it. Renaming a key will break things down the line but not the parser itself. And when OWI eventually breaks some output again - we can patch that specific item within it's parser, while everything else follows the same logic and is easy to remember.

Ulibos:
Status update here: https://github.com/Team-Silver-Sphere/SquadJS/pull/354#discussion_r1566462582
*Sorry for bad placement, my github kung-fu is still weak.*

core/id-parser.js
fantinodavide: I'm good with the new changes after my first review, and with keeping your implementation of the ID parser since you proved to have went through an analysis of all the side effects of both your implementation and the one currently on the `master` branch.

fantinodavide (approved):

squad-server/index.js
fantinodavide: As discussed on Discord with @Ulibos, the best choice appears to be a method that accepts an extra parameter to tune the output based on the needs.

A jsdoc would be a good fit for defining all the possible values for the extra parameter that I will just refer to as "output format parameter".

The output format parameter could be defined with the following jsdoc: `@param {'steamID'|'eosID'|'any'|'player'}` and will return the output formatted as showed in the following example:

- **steamID**: The output stays the same as the current implementation. Example: `[ '01234567891234567', '01234567891234567' ]`
- **eosID**: Example: `[ 'a1b2c3d4e5f6a7b8c9d1e2f3a4b5c6d7e8f9', 'a1b2c3d4e5f6a7b8c9d1e2f3a4b5c6d7e8f9' ]`
- **any**: The eosID should be the preferred ID but include a fallback to steamID in case of missing eosID. Example `[ 'a1b2c3d4e5f6a7b8c9d1e2f3a4b5c6d7e8f9', '01234567891234567' ]`
- **player**: `[ { name: 'example_name', eosID: 'a1b2c3d4e5f6a7b8c9d1e2f3a4b5c6d7e8f9', steamID: '01234567891234567', ... }, { name: 'example_name', eosID: 'a1b2c3d4e5f6a7b8c9d1e2f3a4b5c6d7e8f9', steamID: null|undefined, ... } ]`

fantinodavide:
New changes look good overall. The error handling of the new `getAdminsWithPermission` is very clear and easy to understand.

fantinodavide (approved):
https://github.com/Team-Silver-Sphere/SquadJS/pull/354#issuecomment-2133785533

Ulibos:
I have closed #352 because it is already integrated into this PR so no unsatisfied dependencies left. #347 can be merged in after this PR.

PR #363, opened 2024-07-07
Improve layer information extraction

denizetkar:
# Description
- `SquadServer` regularly invokes `updateLayerInformation` to update the current and next layer variables, namely `this.currentLayer` and `this.nextLayer`.
- This is fetched from `Layers` class, `getLayerByName` method, which filters the pulled layers `this.layers` by the `name` field.
- The layers are fetched from `https://raw.githubusercontent.com/Squad-Wiki/squad-wiki-pipeline-map-data/master/completed_output/_Current%20Version/finished.json`
  - I'm actually a bit concerned that this "current version" is outdated with respect to the latest Squad version 8.0.
- However, the layer name for filtering is fetched from `SquadRcon` class, `getCurrentMap` method which uses the RCON command `ShowCurrentMap`. This command returns an answer like `Current level is Jensen's Range, layer is JensensRange_USA-RGF, factions USA RGF`. The problem here is 2-fold:
  - The regex used to extract the level and the layer names is broken: `^Current level is (.*), layer is (.*)` Because the 2nd regex group actually matches a string like `JensensRange_USA-RGF, factions USA RGF`
  - Even if we correctly extract the "layer name", it actually corresponds to the `layerid` field in the pulled layers

## What this PR does
Solves those 2 problems by:
- Updating the regexes to extract the level and the layer ids from the RCON commands `ShowCurrentMap` and `ShowNextMap` as: `^Current level is ([^,]*), layer is ([^,]*)`
- Changing the matching logic to use the `layerid` field instead of `name`

### DIFF
--- a/squad-server/index.js
+++ b/squad-server/index.js
@@ -500,8 +500,8 @@ export default class SquadServer extends EventEmitter {
       const nextMap = await this.rcon.getNextMap();
       const nextMapToBeVoted = nextMap.layer === 'To be voted';
 
-      const currentLayer = await Layers.getLayerByName(currentMap.layer);
-      const nextLayer = nextMapToBeVoted ? null : await Layers.getLayerByName(nextMap.layer);
+      const currentLayer = await Layers.getLayerById(currentMap.layer);
+      const nextLayer = nextMapToBeVoted ? null : await Layers.getLayerById(nextMap.layer);
 
       if (this.layerHistory.length === 0) {
         this.layerHistory.unshift({ layer: currentLayer, time: Date.now() });
--- a/squad-server/layers/layers.js
+++ b/squad-server/layers/layers.js
@@ -45,8 +45,8 @@ class Layers {
     return null;
   }
 
-  getLayerByName(name) {
-    return this.getLayerByCondition((layer) => layer.name === name);
+  getLayerById(layerId) {
+    return this.getLayerByCondition((layer) => layer.layerid === layerId);
   }
 
   getLayerByClassname(classname) {
--- a/squad-server/rcon.js
+++ b/squad-server/rcon.js
@@ -131,13 +131,13 @@ export default class SquadRcon extends Rcon {
 
   async getCurrentMap() {
     const response = await this.execute('ShowCurrentMap');
-    const match = response.match(/^Current level is (.*), layer is (.*)/);
+    const match = response.match(/^Current level is ([^,]*), layer is ([^,]*)/);
     return { level: match[1], layer: match[2] };
   }
 
   async getNextMap() {
     const response = await this.execute('ShowNextMap');
-    const match = response.match(/^Next level is (.*), layer is (.*)/);
+    const match = response.match(/^Next level is ([^,]*), layer is ([^,]*)/);
     return {
       level: match ? (match[1] !== '' ? match[1] : null) : null,
       layer: match ? (match[2] !== 'To be voted' ? match[2] : null) : null


squad-server/plugins/discord-killfeed.js
Ulibos, line 1: This seems to be an accidental edit but other than that looks good to me. Going to test it soon.

squad-server/plugins/discord-killfeed.js
```diff
@@ -82,11 +82,6 @@ export default class DiscordKillFeed extends DiscordBasePlugin {
           ? `[${info.victim.steamID}](https://steamcommunity.com/profiles/${info.victim.steamID})`
           : 'Unknown',
         inline: true
-      },
-      {
```
werewolfboy13: Why is this removed? Will need to be added back in if you want this PR approved.

werewolfboy13 (changes requested):
Changes are required.

squad-server/plugins/discord-killfeed.js
denizetkar: To quote myself,

> In addition, removes a redundant code in `DiscordKillFeed` class that used to cause an access error, where if `info.victim` was `null` then the field access statement `info.victim.name` would fail, causing SquadJS to crash.

I observed this behavior randomly in my own dedicated server while the code on `master` branch was running. It would throw access error while trying to access `info.victim.name` when `info.victim` was `null`.

After removing this bit, I can confirm that I haven't encountered that error ever since. Also, if you carefully check that list variable, there is already a duplicate object:
```
      {
        name: "Victim's Name",
        value: info.victim ? info.victim.name : 'Unknown',
        inline: true
      },
```

I hope that was enough reason.

werewolfboy13: That isn't a reason to pull it off. It needs to be corrected as there are users that needs to use that info. The crash will need to be corrected elsewhere in the code.

werewolfboy13: Also each change needs to have its own PR. This change isn't tied at all to the main core issue/correction of this PR. That's why I accidentally overlooked it.

denizetkar: done

denizetkar: Although I still do not understand how that is not a duplicate object. The same field is literally defined 2 positions behind except with a value that doesn't fail in case `info.victim` is `null` 🤷 
```typescript
    const fields = [
      {
        name: "Attacker's Name",
        value: info.attacker.name,
        inline: true
      },
      {
        name: "Attacker's SteamID",
        value: `[${info.attacker.steamID}](https://steamcommunity.com/profiles/${info.attacker.steamID})`,
        inline: true
      },
      {
        name: "Attacker's EosID",
        value: info.attacker.eosID,
        inline: true
      },
      {
        name: 'Weapon',
        value: info.weapon
      },
      {
        name: "Victim's Name",
        value: info.victim ? info.victim.name : 'Unknown',
        inline: true
      },
      {
        name: "Victim's SteamID",
        value: info.victim
          ? `[${info.victim.steamID}](https://steamcommunity.com/profiles/${info.victim.steamID})`
          : 'Unknown',
        inline: true
      },
      {
        name: "Victim's Name",
        value: info.victim.name,
        inline: true
      }
    ];
```

Ulibos: While your fix is meaningful in this particular case (I can only assume there was a "bad" merge that caused the duplication) werewolf asks to keep killfeed fix separate from this specific issue simply for clarity. I am sure that if you commit this fix as a separate PR (and providing the rationale from above) it will be merged in.

Ulibos:
I have submitted a PR that addresses duplicate Victim field that you reported: #364

denizetkar:
Hi @Ulibos,

Any updates on testing the PR?

Ulibos:
We are using a plugin that queries data from layers infrequently so I can't vouch for a full comprehensive test but so far I haven't seen a single failure on this part of the code.

denizetkar:
Should we merge it then?

Ulibos:
I'm all for merge but I'm not the one calling the shots.

werewolfboy13:
There is another merge about to happen soon that will require rewrites again. Merges are on hold until then.

werewolfboy13 (approved):

werewolfboy13 (approved):

werewolfboy13:
Force pushing. Mobile UI is preventing conversation resolution.

PR #364, opened 2024-07-11
Remove duplicate "victim" friend from DiscordKillFeed plugin

Ulibos:
Fixes an oopsie I did in EOS migration, duplicating the field in the plugin. I've removed the on that wasn't performing safety checks and was ocasionally crashing people's squadjs.


lbzepoqo:
You're not going to update it to Victim's EOS ID instead just like with the other plugins in #354 ?

Ulibos:
I'm an idiot... I forgot my own code. That is the reason there was that duplicate. It was meant to be rewritten to show EOS which I forgot to do. I'll fix it in a minute.

Ulibos:
Fixed.

PR #366, opened 2024-07-26
feat: add killstreaks plugin

IgnisAlienus:
Adds Killstreaks Plugin that Press Start Gaming has been using for some time now....when we were seeding anyways.


IgnisAlienus (commented):
fix: cases where teamName might be null

PR #370, opened 2024-07-31
Add support for tailing logs over SFTP

Thomas-Smyth:


werewolfboy13 (approved):
Changes look good, will need to monitor if stable on more machines.

PR #379, opened 2024-08-20
Add option to load admin lists via FTP

mbartisan:
The admin-list system now supports loading the Admins.cfg file via FTP, significantly improving compatibility with game server providers. By setting the admin-list configuration type to ftp and providing an FTP URL as the source, the Admins.cfg file will be downloaded into a buffer and parsed by the rest of the system.

```json
    {
        "type": "ftp",
        "source": "ftp://<user>:<password>@<host>:<port>/<url-path>"
      }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,10 @@ The following section of the configuration contains information about your Squad
       {
         "type": "remote",
         "source": "http://yourWebsite.com/Server1/Admins.cfg",
+      },
+      {
+        "type": "ftp",
+        "source": "ftp://<user>:<password>@<host>:<port>/<url-path>",
       }
     ]
   },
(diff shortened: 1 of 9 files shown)


werewolfboy13:
@mbartisan Are we able to provide some documentation in the template folders?

mbartisan:
@werewolfboy13 Yeah I can. I also forgot to commit my removal of the debug logs. I'll update the PR in a sec.

werewolfboy13:
Then please run `yarn build-all`. You will need to have ran `yarn install --ignore-engines` If it is not installed locally yet.

mbartisan:
@werewolfboy13 Latest is pushed. Lmk if there are any other docs that should get updated.

werewolfboy13:
There is, see above. May want to lint as well `yarn lint`

mbartisan:
@werewolfboy13 Updated. The build and linter modified some files that were outside of my changes. They're in separate commits for each if any should get culled.

werewolfboy13 (approved):

werewolfboy13:
@mbartisan that is fine. I always do squash and merge.

PR #385, opened 2024-10-13
Support for fetching admin list from SFTP source

pavelhoral:
This PR adds support for fetching Admins.cfg file from SFTP source. I have also replaced manual regex based URL parsing with NodeJS's URL parsing.

I have tested both FTP and SFTP using local docker containers (https://github.com/garethflowers/docker-ftp-server and https://docs.sftpgo.com/).

I have used `ssh2-sftp-client` package because this exact package is being used by the current ftp-tail dependency (https://github.com/Thomas-Smyth/ftp-tail/). The major version is increased from 10 to 11, but I don't think that should break anything.


werewolfboy13 (changes requested):
Apologies on the delay, this needs to be corrected to use the templates.

pavelhoral:
> Apologies on the delay, this needs to be corrected to use the templates.

I am not sure I follow what needs to be done.

Hans-Vader:
@pavelhoral @werewolfboy13 is there a way to finish this PR? This seems to be a reasonable addition.

Issue #386, opened 2024-10-26
[SEV 2] Layers File broken on recent update


werewolfboy13:
### Description of Issue

Squad JS needs to updated ASAP to support the new layer format. Will need support to patch this.

### Errors or Screenshots of Issue

![image](https://github.com/user-attachments/assets/a0699644-b8ec-4b39-9a64-ded9d53a6621)

Effected PRs: https://github.com/Squad-Wiki/squad-wiki-pipeline-map-data/pull/116 and https://github.com/Squad-Wiki/squad-wiki-pipeline-map-data/pull/115.

### Squad Information
 * SquadJS Version: Current as of Writing
 * Squad Version: Current as of Writing

If potentially relevant, please provide regarding the state of the Squad server at the time of error, e.g. the current layer.

### System Information
 * Operating System:
 * NodeJS Version:
 * Yarn Version:

watchful-eyes:
The fix will not cover the current data we get from the last version since some data is gone in the 8.1. Such as, vehicle list doesn't exist, name/faction will be the same etc.

Fix PR:
https://github.com/Team-Silver-Sphere/SquadJS/pull/387

PR #387, opened 2024-10-26
[SEV 2] Layers File Fix for recent update

watchful-eyes:
Fix for https://github.com/Team-Silver-Sphere/SquadJS/issues/386

### DIFF
--- a/squad-server/layers/layer.js
+++ b/squad-server/layers/layer.js
@@ -20,23 +20,11 @@ export default class Layer {
     for (const t of ['team1', 'team2']) {
       let team = data[t] || data.teamConfigs[t];
       this.teams.push({
-        faction: team.faction,
-        name: team.teamSetupName,
+        faction: team.defaultFactionUnit, 
+        name: team.defaultFactionUnit, // There is no other name presented in the data so in order to keep the same structure to support the old formatted codes, used the same name
         tickets: team.tickets,
-        commander: team.commander,
-        vehicles: (team.vehicles || []).map((vehicle) => ({
-          name: vehicle.type,
-          classname: vehicle.rawType,
-          count: vehicle.count,
-          spawnDelay: vehicle.delay,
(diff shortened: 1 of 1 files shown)


werewolfboy13 (changes requested):
Please provide logs from `https://raw.githubusercontent.com/Squad-Wiki/squad-wiki-pipeline-map-data/dev/completed_output/_Current%20Version/finished.json` showing proper loading to ensure this passes. If so I'll push and bump the RC1 release to prod. @Shanomac99 so you can find the git versioning corrections. 

If all passes will be pushed when I wake back up.

watchful-eyes:
image
image

don't know what you need exactly so hope these should suffice

werewolfboy13 (approved):

werewolfboy13:
Team don't push yet please.

squad-server/layers/layer.js
```diff
@@ -20,23 +20,11 @@ export default class Layer {
...
       let team = data[t] || data.teamConfigs[t];
       this.teams.push({
-        faction: team.faction,
-        name: team.teamSetupName,
+        faction: team.defaultFactionUnit, 
+        name: team.defaultFactionUnit, // There is no other name presented in the data so in order to keep the same structure to support the old formatted codes, used the same name
         tickets: team.tickets,
-        commander: team.commander,
-        vehicles: (team.vehicles || []).map((vehicle) => ({
-          name: vehicle.type,
```
werewolfboy13, line 27: Isn't there some custom plugins that use this data still?

watchful-eyes: I think dave's mapvote uses to show what the layer have as vehicles in the broadcast. But since the new layer list doesn't provide this info anymore we can't get that

watchful-eyes: best thing for custom plugins would be set it as empty array

PR #388, opened 2024-11-05
feat: allow execution of raw RCON commands via WebSocket clients

anklav24:


werewolfboy13 (commented):
@Thomas-Smyth you good with this? If the changes won't break anything.

watchful-eyes:
> @Thomas-Smyth you good with this? If the changes won't break anything.

Related PR:
https://github.com/Team-Silver-Sphere/SquadJS/pull/205

anklav24:
> > @Thomas-Smyth you good with this? If the changes won't break anything.
> 
> Related PR: #205

Got it! I actually like this option with `super` even more. The issue here is that the Socket IO plugin does not expose to use methods from the mail class, such as execute.

JiaruiGe:
May I ask if it is possible to use rcon to send native commands after connecting to socketio in an external program after this modification, such as AdminForceTeamChange id

lbzepoqo:
> May I ask if it is possible to use rcon to send native commands after connecting to socketio in an external program after this modification, such as AdminForceTeamChange id

With this pull request, you can

PR #394, opened 2025-01-31
Fix DLC match group in new-game.js

CamperNoob:
NEW_GAME wasn't triggered when map was changed to modded one, since modded map's path can also include numbers, dashes and underscores.
Example:
`[2025.01.31-09.45.27:921][378]LogWorld: Bringing World /TitanLeagueMod_44th/Maps/Anvil/Gameplay_Layers/TL_Anvilv1.TL_Anvilv1 up for play (max tick rate 40) at 2025.01.31-11.45.27`


werewolfboy13 (changes requested):
Why does this have so many commits not related to what the PR is talking out?

CamperNoob:
Wasn't the case exactly 2 hours before the review, lmao. I'll do a cherry-pick and PR from separate branch.

PR #400, opened 2025-02-10
Added new plugin to enforce squadleader take SL kit

Acurasa:
The **EnforceSquadLeaderKit** plugin was created to maintain structured and effective team gameplay in Squad servers. It ensures that squad leaders comply with the requirement of selecting the proper Squad Leader kit, a role essential for efficient communication and coordination during matches.

### **Purpose and Features**

This plugin automatically monitors squad leaders and provides the following functionalities:

- **Warning System:** Issues configurable warnings when squad leaders do not have the correct Squad Leader kit.
- **Automatic Squad Disbanding:** If a squad leader ignores warnings and fails to switch to the appropriate kit, the plugin disbands their squad.
- **Warning Reset:** Warnings are cleared if the player selects the correct kit after receiving a notice.

### **Why Use This Plugin?**

- **Enhanced Team Coordination:** Ensuring squad leaders have the right kits improves communication and combat efficiency.
- **Automated Monitoring:** Reduces the administrative workload for server moderators by automating checks and compliance enforcement.
- **Configurable Behavior:** Offers flexibility through customizable warning messages, intervals, and thresholds.

This plugin was designed to encourage responsible gameplay and maintain a positive experience for all players by enforcing squad leader role compliance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -786,7 +786,55 @@ Grafana:
            <h6>Default</h6>
            <pre><code>false</code></pre></li></ul>
         </details>
+<details>
+  <summary>EnforceSquadLeaderKit</summary>
+  <h2>EnforceSquadLeaderKit</h2>
+  <p>The <code>EnforceSquadLeaderKit</code> plugin ensures that squad leaders select the correct Squad Leader kit. It will warn players who do not comply and disband their squad if they fail to switch after multiple warnings.</p>
 
+  <h3>Options</h3>
+  <ul>
+    <li>
+      <h4>warningMessage</h4>
+      <h6>Description</h6>
+      <p>Message displayed to squad leaders warning them to take the correct kit.</p>
+      <h6>Default</h6>
+      <pre><code>You must take the Squad Leader kit, or your squad will be disbanded!</code></pre>
+    </li>
+
+    <li>
+      <h4>checkInterval</h4>
+      <h6>Description</h6>
+      <p>Interval (in seconds) to check all squad leaders for the correct kit.</p>
+      <h6>Default</h6>
+      <pre><code>30</code></pre>
+    </li>
+
+    <li>
+      <h4>maxWarnings</h4>
+      <h6>Description</h6>
+      <p>Number of warnings a squad leader receives before their squad is disbanded.</p>
+      <h6>Default</h6>
+      <pre><code>3</code></pre>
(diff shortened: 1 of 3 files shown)


fantinodavide (changes requested):
I believe the SquadJS setup wasn't properly done, causing the pre-commit hooks not to run. The README.md file seems to have been manually updated and the changes do not respect the original format.
Skipping the pre-commit hooks would cause changes to the config.json and README.md in future commits, potentially unrelated.

werewolfboy13:
The have to be done in the template folders and build actions need to be done.

watchful-eyes:
I have a few suggestions for the plugin before pushing into the prod. 
- It'd be better if you check the "seed" status of the server before checking Squad Leaders. 
- Pilot, Crewman kits don't include "SL" in their kit name also there are some kits that called as "SquadLeader" instead of "SL" afaik so, please be sure of the kit names too.

fantinodavide:
@watchful-eyes 
> * It'd be better if you check the "seed" status of the server before checking Squad Leaders.

I'd also check if the server is on the training range

> * Pilot, Crewman kits don't include "SL" in their kit name also there are some kits that called as "SquadLeader" instead of "SL" afaik so, please be sure of the kit names too.

Vanilla roles all have "SL" as part of the name. Examples: "SLCrewman" and "SLPilot".

byander:
I have other suggestions:

* Send a warning to all members of the squad;
* Maybe put an option time for the player to change the role, because sometimes the SL quits the game and the HAB doesn't have ammo to player change to the SL kit.

Good job on this plugin!

lbzepoqo:
> I have other suggestions:
> 
>     * Send a warning to all members of the squad;
> 
>     * Maybe put an option time for the player to change the role, because sometimes the SL quits the game and the HAB doesn't have ammo to player change to the SL kit.
> 
> 
> Good job on this plugin!

I like this idea since it also informs the members that the squad is likely to be automatically disbanded instead of being booted out from the squad without knowing the reason. Some even complains why they got kicked from the squad. At least they can anticipate whether they will be disbanded or someone will step up to take the lead.

H4wKs:
It would be nice to be able to customise this message : "Thank you for taking the correct kit!" in the plugin config.

PR #402, opened 2025-04-26
Revert "Fix Regex detection for new games for modded layers"

werewolfboy13:
Reverts Team-Silver-Sphere/SquadJS#401


werewolfboy13:
Testing a fix on the prod workflow.

PR #404, opened 2025-06-04
layers: Fetch a more up-to-date layers list

zagor:
The layers list fetched in master is old and does not reflect the current game.

PR #405, opened 2025-06-15
Use modern layers.json

### LINKED ISSUE #386: [SEV 2] Layers File broken on recent update

zagor:
Use the newer layers.json file which contains modern game data.

Look up layer factions and units in updateServerInformation().

### DIFF
--- a/squad-server/index.js
+++ b/squad-server/index.js
@@ -204,6 +204,7 @@ export default class SquadServer extends EventEmitter {
       this.layerHistory = this.layerHistory.slice(0, this.layerHistoryMaxLength);
 
       this.currentLayer = data.layer;
+      await this.updateServerInformation();
       await this.updateAdmins();
       this.emit('NEW_GAME', data);
     });
@@ -534,6 +535,19 @@ export default class SquadServer extends EventEmitter {
     return this.updateServerInformation();
   }
 
+  setupTeams() {
+    for (const t in [0, 1]) {
+      const unit = Layers.units[this.unitNames[t]];
+      this.currentLayer.teams[t] = {
+        faction: unit.factionID,
+        name: unit.displayName,
+        tickets: this.currentLayer.tickets[t],
+        commander: this.currentLayer.commander,
+        vehicles: unit.vehicles,
+      };
+    }
+  }
+
   async updateServerInformation() {
     if (this.updateA2SInformationTimeout) clearTimeout(this.updateA2SInformationTimeout);
 
@@ -543,7 +557,7 @@ export default class SquadServer extends EventEmitter {
       const rawData = await this.rcon.execute(`ShowServerInfo`);
       Logger.verbose('SquadServer', 3, `Server information raw data`, rawData);
       const data = JSON.parse(rawData);
-      Logger.verbose('SquadServer', 2, `Server information data`, JSON.data);
+      Logger.verbose('SquadServer', 2, `Server information data`, data);
 
       const info = {
         raw: data,
@@ -561,8 +575,8 @@ export default class SquadServer extends EventEmitter {
         currentLayer: data.MapName_s,
         nextLayer: data.NextLayer_s,
 
-        teamOne: data.TeamOne_s?.replace(new RegExp(data.MapName_s, 'i'), '') || '',
-        teamTwo: data.TeamTwo_s?.replace(new RegExp(data.MapName_s, 'i'), '') || '',
+        teamOne: data.TeamOne_s,
+        teamTwo: data.TeamTwo_s,
 
         matchTimeout: parseFloat(data.MatchTimeout_d),
         matchStartTime: this.getMatchStartTimeByPlaytime(data.PLAYTIME_I),
@@ -579,6 +593,7 @@ export default class SquadServer extends EventEmitter {
       this.playerCount = info.playerCount;
       this.publicQueue = info.publicQueue;
       this.reserveQueue = info.reserveQueue;
+      this.unitNames = [data.TeamOne_s, data.TeamTwo_s];
 
       this.matchTimeout = info.matchTimeout;
       this.matchStartTime = info.matchStartTime;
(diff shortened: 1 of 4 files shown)


zagor:
This PR replaces #404.

fantinodavide:
Haven't had a chance to test these changes yet, but I went through the whole diffs and the changes are looking good to me

werewolfboy13:
I would need a server to test this to make sure no breaking changes are occuring.

zagor:
There are breaking changes, for a strict definition of breaking.

The new game data file contains data that the old file doesn't, and vice versa. Any plugin depending on that old deprecated data would therefore break. But since we (or at least I) don't know what custom plugins people use there is no way to know the scope of this hypothetical breakage.

Also, the actual live game servers have moved away from this old data format since quite a long time (over a year?), which would already have broken any plugins depending on it. In that context, I think it is unlikely this data format change breaks any plugin that currently works correctly against master.

werewolfboy13:
Any changes need testing on live servers before being pushed to prod. Will only push once we have a validated test. 

This is from lessons learned from other incidents.

zagor:
> Any changes need testing on live servers before being pushed to prod. Will only push once we have a validated test.

I agree completely. I have only tested this on my test server so far, not a live server. I am in the process of deploying it on a live server "soon", hopefully in the next couple of weeks.

Ideally several people should try it so we don't rely on testing only by the PR author.

zagor:
Ok, this is now running on a live server since a few days. The only issue I've found was that immediately after game server restart, we get one or two server information blocks without any teams defined. So I've added a check for that.

zagor:
I have found that this solution is not good enough for my needs.

The information returned by rcon command `ShowServerInfo` is up to 30 seconds old. Maybe this is old news, but it was unknown to me. The game server only updates this data when it posts a server information update to EOS. The command is, essentially, returning the server information that is shown in the server browser. The effect of this is that the function returns information about the previous round up to 30 seconds into the staging of a new round, causing problems for any plugin looking for accurate information.

I cannot therefore update teams and vehicle information in `updateServerInformation()` but must instead do it in `updateLayerInformation()`, which does gets proper accurate server status.

However, the `ShowCurrentMap` command used by `updateLayerInformation()` returns team information in the form `ADF+Mechanized` rather than the `ADF_LO_Mechanized` format that we get from `ShowServerInfo`, and which we need to look up units in the layer file. So I need to do a little transformation from one form to the other, in order to get accurate unit information at the start of a round.

Background for this effort: The reason I need team information timely is that I have written a plugin for automatic enforcement of vehicle claim rules. It needs to know which vehicles are available on each side at the start of the round when the race for claims occurs.

zagor:
### Change how layer and factions are discovered

Since the data returned by `ShowServerInfo` is delayed up to 30 seconds, it can not be relied on to determine current layer and faction. Especially not at the start of rounds. Instead, use `ShowCurrentMap` and `ShowNextMap` for this. They return the actual current state of the server.

Also, since the same layer object can be returned for both the current map and the next, we cannot store team information there. Thus, a new team object is returned in addition to the layer object. This requires parsing and returning additional fields from `rcon.getCurrentMap()` and `rcon.getNextMap()`.

zagor:
(Rebased on master)

zagor:
This is running on server "Squad International", where it forms the foundation for an automatic vehicle claims handling plugin. 

Anyone interested in seeing it in action can join the server and/or ping me. I'm "Baxter" in Squad and Discord.

werewolfboy13 (dismissed):

werewolfboy13 (changes requested):
PR title needs to be corrected or separate PRs need to be made for the additional changes.

zagor:
Thank you for the feedback. Can you clarify what you consider to be additional changes and what would then be the core changes?

werewolfboy13:
> ### Change how layer and factions are discovered
> 
> 
> 
> Since the data returned by `ShowServerInfo` is delayed up to 30 seconds, it can not be relied on to determine current layer and faction. Especially not at the start of rounds. Instead, use `ShowCurrentMap` and `ShowNextMap` for this. They return the actual current state of the server.
> 
> 
> 
> Also, since the same layer object can be returned for both the current map and the next, we cannot store team information there. Thus, a new team object is returned in addition to the layer object. This requires parsing and returning additional fields from `rcon.getCurrentMap()` and `rcon.getNextMap()`.

This here is not in the title. Each core change needs to be its own PR.

werewolfboy13:
This will satisfy #386 as an interim fix.

zagor:
Ok, coming back to this now.

@werewolfboy13 Do you require that they are separate PRs or is it enough that they are separate commits in this PR?

Making two PRs means first making a PR that works with the old layer code and team objects. And then a second PR that substantially rewrites what was in the first PR. The code in the first PR will thus only live until the second is merged, which seems a little wasteful. But I can do it if you really prefer it that way.

lbzepoqo:
I'm getting this

```
(node:36) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of undefined (reading 'tickets')
    at new Layer (file:///home/container/squad-server/layers/layer.js:21:44)
    at Layers.pull (file:///home/container/squad-server/layers/layers.js:29:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async SquadServer.watch (file:///home/container/squad-server/index.js:72:5)
    at async main (file:///home/container/index.js:17:3)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:36) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
```

PR #406, opened 2025-06-22
Update rcon.js

20R-Grumpus:
updates to getCurrentMap() and getNextMap() and helps fix layer names returned as unknown

Null-safe fallback 
Improved regex | Handles unexpected trailing commas or extra info after the layer name.
Layer name cleanup | Treats "unknown" or empty strings as invalid layer values.
Returns layerid and name 

includes logging for layer response info that may be removed

### DIFF
--- a/squad-server/rcon.js
+++ b/squad-server/rcon.js
@@ -131,18 +131,53 @@ export default class SquadRcon extends Rcon {
 
   async getCurrentMap() {
     const response = await this.execute('ShowCurrentMap');
-    const match = response.match(/^Current level is ([^,]*), layer is ([^,]*)/);
-    return { level: match[1], layer: match[2] };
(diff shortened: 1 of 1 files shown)


20R-Grumpus:
Layer name returned as Unknown still occurs but can be handled within plugins to keep asking

werewolfboy13:
Rejecting in favor of #405. We need QA testing to have a valid path.

20R-Grumpus:
hopefully #405 solves similar issues, i am not fully up to speed on those changes. Was just asked to submit by dave. The purpose of this is to improve functionality for discord-server-status.js regularly failing to show the layer whilst at the same time preventing other custom plugins from working in the event that the map is returned as unknown. 

However this change has been deployed on our live server since a few days prior to this PR and has not resulted in any breakage beyond when discord-server-status.js receives a layer name that includes underscores. In which case it wont load the map picture, but the discord bots status and its embed never show unknown.

PR #407, opened 2025-07-11
Show reserved slots in the DiscordServerStatus plugin

zagor:
Update discord status line to include reserved slots in player count.

### DIFF
--- a/squad-server/plugins/discord-server-status.js
+++ b/squad-server/plugins/discord-server-status.js
@@ -123,8 +123,15 @@ export default class DiscordServerStatus extends DiscordBaseMessageUpdater {
   async updateStatus() {
     if (!this.options.setBotStatus) return;
 
+    let players = this.server.a2sPlayerCount;
+    if (this.server.publicQueue || this.server.reserveQueue)
+      players += `+${this.server.publicQueue + this.server.reserveQueue}`;
+
+    let slots = this.server.publicSlots;
+    if (this.server.reserveSlots) slots += `+${this.server.reserveSlots}`;
+
     await this.options.discordClient.user.setActivity(
-      `(${this.server.a2sPlayerCount}/${this.server.publicSlots}) ${
+      `(${players}/${slots}) ${
         this.server.currentLayer?.name || 'Unknown'
       }`,
       { type: 4 }


squad-server/plugins/discord-server-status.js
```diff
@@ -123,8 +123,15 @@ export default class DiscordServerStatus extends DiscordBaseMessageUpdater {
...
+    let players = this.server.a2sPlayerCount;
+    if (this.server.publicQueue || this.server.reserveQueue)
+      players += `+${this.server.publicQueue + this.server.reserveQueue}`;
+
+    let slots = this.server.publicSlots;
+    if (this.server.reserveSlots) slots += `+${this.server.reserveSlots}`;
+
     await this.options.discordClient.user.setActivity(
-      `(${this.server.a2sPlayerCount}/${this.server.publicSlots}) ${
+      `(${players}/${slots}) ${
```
werewolfboy13, line 134: Recommended that this follow the rest of the repo. Ex making it `this.server.players` etc. To keep it consistent among the rest of repo.

werewolfboy13 (dismissed):

squad-server/plugins/discord-server-status.js
zagor: I'm not sure I follow. Do you mean we should use `this.server.playerCount` instead of `this.server.a2sPlayerCount`?

I am mimicking the method `generateMessage()` in this file which uses `this.server.a2sPlayerCount`. Should we change all uses of `a2sPlayerCount` in this file to `playerCount`?

werewolfboy13: @Thomas-Smyth has a prefered setup for it, i'll wait for him to chime in.

Thomas-Smyth: I agree with the pull request as is - it is consistent with the remainder of the plugin at least and I believe the values used are slightly more reliable that `this.servers.players` at times, e.g. during map changes.

Thomas-Smyth (approved):
Looks good to me. Although, how's it look in Discord? Does the additional text fit in the user list well?

zagor:
It only adds 5 characters at most, so it fits well. I'll post a screenshot when the server fills up later today.

zagor:
Here's an example:
image

PR #408, opened 2025-07-13
Refactor SquadLogParser to dynamically load parser files

bombitmanbomb:
Refactored the log parser to find and load all parsers in log-parser folder on startup.

This allows custom log parsers and events to be dropped-in for a variety of things, mainly Mods.
Changes stemmed from needing some custom parsers for the [King of the Hill](https://steamcommunity.com/sharedfiles/filedetails/?id=3443779816) gamemode, which has some special logging to trigger events.

Tested on `(AFO) Noobs & Boobs`

image
image

### DIFF
--- a/squad-server/log-parser/index.js
+++ b/squad-server/log-parser/index.js
@@ -2,7 +2,7 @@ import fs from 'fs';
 import path from 'path';
 
 import LogParser from 'core/log-parser';
-import Logger from 'core/logger.js';
+import Logger from 'core/logger';
 import { fileURLToPath, pathToFileURL } from 'url';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
--- a/squad-server/log-parser/index.js
+++ b/squad-server/log-parser/index.js
@@ -11,7 +11,12 @@ export default class SquadLogParser extends LogParser {
   constructor(options) {
     super('SquadGame.log', options);
     this._rules = [];
-    this.setupRules();
+  }
+
+  async watch() {
+    //? Wait for rules to be configured before hooking the log file
+    await this.setupRules();
+    return super.watch();
   }
 
   async setupRules() {
--- a/squad-server/log-parser/index.js
+++ b/squad-server/log-parser/index.js
@@ -2,7 +2,7 @@ import fs from 'fs';
 import path from 'path';
 
 import LogParser from 'core/log-parser';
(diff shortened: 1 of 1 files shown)


werewolfboy13:
This repo meets QA guidelines, awaiting confirmation on some other members since there is a medium risk of breaking existing setups.

werewolfboy13:
Code meets standards and does not flag on `yarn lint`.
image

werewolfboy13 (approved):

Thomas-Smyth:
Is there any check to ensure that the rules have been loaded prior to any logs being parsed? Currently, I cannot see any which would mean that potentially logs could be processed before they're loaded leading to things not working as expected. 

I think the rules need to be loaded either 1) when initialising the log parser (which would be more difficult as that's in a non-asynchronous function), 2) where `.watch()` is called, 3) in the `.watch()` function. 3 seems like the best option and would only require the `.setupRules()` to be called and awaited in the `.watch()` method. It may require the parent class to be told where to load the rules from.

bombitmanbomb:
The load can't be done synchronously due to `import()`, but I will look at what it will take to do in `watch()`

bombitmanbomb:
Okay this is actually really easy to do, adding it now

bombitmanbomb:
Added an override to the `watch()` function which now loads the log files and then calls the underlying LogParser's watch function
image

bombitmanbomb:
It would appear this doesnt work in Windows due to some ESM import rules requiring an exact file:/// path

bombitmanbomb:
Fixed by using resolve's and pathToFileURL, which works on POSIX & Windows import

Thomas-Smyth:
It appears you have reverted the changes to the `watch` method by accident in 5e2b74e966532512ce85fd2985e3adbaec61201a.

werewolfboy13:
@bombitmanbomb can you fix or see what the issue is. Can't approve this PR without it.

werewolfboy13 (changes requested):
Comments on PR need resolving.

bombitmanbomb:
@Thomas-Smyth Some weird file desync w/ github broke stuff. Should be fixed now.

lbzepoqo:
I will try to run this in our server

EDIT: No issues so far

image

werewolfboy13:
Bypass applied as this PR is not loading correctly on my end. Approved PR.

PR #410, opened 2025-07-20
Improve format of Community Ban List Discord alerts

Makavich:
it now SHOWS SteamID and EOS ID 

image

### DIFF
--- a/squad-server/plugins/cbl-info.js
+++ b/squad-server/plugins/cbl-info.js
@@ -126,7 +126,7 @@ export default class CBLInfo extends DiscordBasePlugin {
             },
             {
               name: 'Reputation Points',
-              value: `${data.steamUser.reputationPoints}`,
+              value: `${data.steamUser.reputationPoints} (${data.steamUser.reputationPointsMonthChange || 0} from this month)`,
               inline: true
             },
             {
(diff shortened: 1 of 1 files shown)


squad-server/plugins/cbl-info.js
```diff
@@ -117,11 +114,19 @@ export default class CBLInfo extends DiscordBasePlugin {
...
+              name: 'EOS ID',
+              value: info.player.eosID || 'Unknown',
+              inline: true
+            },
             {
               name: 'Reputation Points',
-              value: `${data.steamUser.reputationPoints} (${
-                data.steamUser.reputationPointsMonthChange || 0
-              } from this month)`,
+              value: `${data.steamUser.reputationPoints}`,
```
Thomas-Smyth: Is there a particular reason for removing the amount of points from the current month? That seems like a useful statistic to establish how recent their poor behaviour is.

Thomas-Smyth (commented):
Looks good! Please see the point below.

squad-server/plugins/cbl-info.js
Makavich: you may add it , it was removed by ai

Thomas-Smyth: Please could you apply that change? Unfortunately, it's not easy for me to edit other's pull requests.

Thomas-Smyth (approved):

squad-server/plugins/cbl-info.js
Thomas-Smyth: Never mind, I managed it somehow.

PR #413, opened 2025-09-20
Fix player-disconnected log regex for 9.0

zagor:
Squad 9.0 did some changes in the log message.

Only require the things we care about so it is more robust against future changes.


watchful-eyes:
Tested in multiple live servers, works.

werewolfboy13 (approved):

PR #417, opened 2026-04-15
fix(team-randomizer): add safety improvements

lbzepoqo:
## Summary

Several safety and UX issues in `TeamRandomizer` that affect production servers:

- **Disabled by default** — changed `defaultEnabled` from `true` to `false`. This is a destructive, irreversible-during-round operation; it shouldn't be on without explicit opt-in.
- **Re-entrancy guard** — concurrent `!randomize` calls from multiple admins would run simultaneously and clobber each other. Added a `this.randomizing` flag with `try/finally` to ensure it always resets even on error.
- **Error handling per player** — the original loop had no try/catch; a single failed `switchTeam` RCON call would abort mid-shuffle, leaving teams in a partially-randomized state. Now each switch is caught individually, logs the failure, and continues.
- **Player broadcast** — players were silently moved to different teams with no explanation. Added an `AdminBroadcast` before switching starts.
- **Admin feedback** — the admin who ran the command got no confirmation. Added a `warn` on completion with the count of players switched.
- **Shuffle cleanup** — removed the unnecessary `temporaryValue` temp variable; replaced with destructuring swap.

## Test plan

- [ ] Run `!randomize` in admin chat — confirm broadcast fires, teams are shuffled, admin receives completion message
- [ ] Run `!randomize` twice simultaneously — confirm second call is rejected with "already in progress" message
- [ ] Verify plugin is opt-in by default (not enabled in generated config)

### DIFF
--- a/squad-server/plugins/team-randomizer.js
+++ b/squad-server/plugins/team-randomizer.js
@@ -9,7 +9,7 @@ export default class TeamRandomizer extends BasePlugin {
   }
 
   static get defaultEnabled() {
-    return true;
+    return false;
   }
 
   static get optionsSpecification() {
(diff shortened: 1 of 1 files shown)


fantinodavide (changes requested):
@lbzepoqo the plugin file itself looks good, but the pre-commit hook didn't run, which is a mandatory step for changes like this one.

lbzepoqo (commented):
Pre-commit hook steps have now been applied: lint (ESLint + Prettier) reformatted `team-randomizer.js`, and `yarn build-all` regenerated `README.md` and `config.json`. Sorry for the oversight!

PR #418, opened 2026-04-16
chore: rebuild README.md and config.json

lbzepoqo:
## Summary

- Regenerates `README.md` and `config.json` via `yarn build-all`
- Pure ordering/formatting diff — no content changes to any plugin documentation
- Separates the docs regeneration from plugin-specific PRs to keep histories clean

## Test plan

- [ ] Verify no plugin documentation content was added or removed
- [ ] Confirm `yarn build-all` produces no further diff after merging


fantinodavide:
@lbzepoqo I believe the issue should be resolved at the root, adding a proper sorting in the methods that do the actual generation of the config.json file in https://github.com/Team-Silver-Sphere/SquadJS/blob/master/squad-server/factory.js

lbzepoqo:
Fixed in 938f70c. The sort comparators in both `buildConfig()` and `buildReadmeFile()` were comparing `a.name` / `b.name` — but `Object.keys()` returns strings, not plugin objects, so `.name` was always `undefined` and the sort was a no-op. Changed to compare the key strings directly (`a < b`), then regenerated docs with the now-deterministic ordering.

PR #420, opened 2026-05-05
TeamRandomizer: strict command match & code style

mikebjoyce:
Enforce an exact, trimmed, case-insensitive chat command match to prevent accidental team randomization when admins mention the command word in normal chat (e.g., "I'm going to randomize the teams"). Adds an explanatory comment and guard check. Also applies stylistic cleanups and formatting: convert quotes to double quotes, add trailing commas, and reformat argument/await lines for readability.

### DIFF
--- a/squad-server/plugins/team-randomizer.js
+++ b/squad-server/plugins/team-randomizer.js
@@ -1,10 +1,10 @@
-import BasePlugin from './base-plugin.js';
+import BasePlugin from "./base-plugin.js";
 
 export default class TeamRandomizer extends BasePlugin {
   static get description() {
     return (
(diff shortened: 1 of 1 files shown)


mikebjoyce:
The accidental trigger occurs in admin chat specifically, not normal player chat, which is already blocked by an earlier guard. This fix ensures only a bare !randomize message in admin chat triggers the shuffle, not sentences that happen to contain the command word.

Gractus:
I think the real fix needed here is adding an anchor to the regex before the CHAT_COMMAND event is emitted.

i.e.

`!([^ ]+) ?(.*)` -> `^\s*!([^ ]+) ?(.*)`

I can't think of any instance where someone would expect a command to be executed when it's included in a sentence.

